PR #13, opened 2022-04-06
Cleanup span in persistent middleware fixes

### LINKED ISSUE #12: Fix span cleanup in persistent middleware
https://github.com/iand675/hs-opentelemetry/blob/4d3666aa8cebb9b3200b84e7ec47c60eb145c771/instrumentation/persistent/src/OpenTelemetry/Instrumentation/Persistent.hs#L76

The context is not properly restored when no parent span was present. A similar issue was present for inSpan and resolved by d9f25cb5bce8867265fc05baee5397e0dd2a1cfa I can submit a pull request with a similar fix.

jfraudeau:
Fixes  #12


iand675:
Thanks for the fix!

PR #15, opened 2022-05-06
faster CI

kakkun61:
- add a cache step to the CI build


iand675:
Thank you for the PR!

PR #24, opened 2022-08-08
Add hspec instrumentation

lf-:
It would be useful to have instrumentation of test suites. My
original use case for this is to validate a change in Persistent
instrumentation, for which hitting it really hard with the Mercury test
suite seems like a reasonable strategy.

### DIFF
--- a/examples/hspec/.gitignore
+++ b/examples/hspec/.gitignore
@@ -1,0 +1,4 @@
+*.aes
+*.hi
+*.o
+*.cabal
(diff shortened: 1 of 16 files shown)


iand675 (changes requested):
@lf- I'm very happy to add this to the repo, but can you add some docstrings and some content to the README?

lf-:
Yep, done :)

iand675 (approved):

PR #33, opened 2022-09-28
Fix compilation error for cabal users

kakkun61:
No code changes too

- proto-lens 0.7.1.2 already supports GHC 9.4
    - https://github.com/google/proto-lens/pull/440
- adds C libraries to _shell.nix_
- adds a CI job using Cabal
- updates the hie.yaml
- updates cabal files
    - runs hpack

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,99 @@
+CABALS := \
+  api/hs-opentelemetry-api.cabal\
+  examples/hspec/hspec-example.cabal\
+  examples/yesod-minimal/yesod-minimal.cabal\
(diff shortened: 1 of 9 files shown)


iand675 (approved):

Makefile
```diff
@@ -0,0 +1,99 @@
+CABALS := \
```
iand675, line 1: I agree that we should have a script to run hpack on `package.yaml` files, but this seems like it would be simpler as a shell strict.

kakkun61: Yes, maybe we will forget to add new cabal files here.

PR #35, opened 2022-09-28
Format codes

### LINKED ISSUE #29: Code Formatter
I think formatting codes with a code formatter is better.

- Introduce a code formatter?
- Which code formatter?
  - [stylish-haskell](https://github.com/haskell/stylish-haskell)
  - [ormolu](https://github.com/tweag/ormolu)
  - [fourmolu](https://github.com/fourmolu/fourmolu)
  - etc.

kakkun61:
- formatting codes
  - I moved some CPP `#if`s, because fourmolu cannot treat them.
- CI for check if codes are formatted
- Make recipes

Closes #29

### DIFF
--- a/fourmolu.yaml
+++ b/fourmolu.yaml
@@ -1,0 +1,8 @@
+indentation: 2
+comma-style: leading # 'leading' or 'trailing'
+record-brace-space: true # rec {x = 1} vs. rec{x = 1}
+indent-wheres: false
+diff-friendly-import-export: false # 'false' uses Ormolu-style lists
+respectful: false # don't be too opinionated about newlines etc.
+haddock-style: multi-line
+newlines-between-decls: 2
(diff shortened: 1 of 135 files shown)


fourmolu.yaml
```diff
@@ -0,0 +1,8 @@
+indentation: 2
+comma-style: leading # 'leading' or 'trailing'
+record-brace-space: true # rec {x = 1} vs. rec{x = 1}
+indent-wheres: false
```
iand675, line 4: `indent-wheres: false` didn't do what I expected. Can we change it to `true` before merging this PR?

iand675 (commented):
@kakkun61 thank you for your hard work. I changed my mind about one formatting option, but this otherwise looks good to me.

fourmolu.yaml
kakkun61: Yes, done!

iand675 (approved):

PR #37, opened 2022-10-05
add instances for monad-tracer

kakkun61:
To add instances for monad-tracer because of utility.

### DIFF
--- a/api/hs-opentelemetry-api.cabal
+++ b/api/hs-opentelemetry-api.cabal
@@ -83,6 +83,7 @@ library
     , template-haskell
     , text
     , thread-utils-context ==0.2.*
+    , transformers
     , unliftio-core
     , unordered-containers
     , vault
@@ -122,6 +123,7 @@ test-suite hs-opentelemetry-api-test
     , template-haskell
     , text
     , thread-utils-context ==0.2.*
+    , transformers
     , unliftio-core
     , unordered-containers
     , vault
--- a/api/package.yaml
+++ b/api/package.yaml
@@ -36,6 +36,7 @@ dependencies:
 - clock
 - memory
 - mtl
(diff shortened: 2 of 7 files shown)


iand675:
I'm not sure if most of the instances here really work. 

The variants of `inSpan`, `inSpan'`, `inSpan''` in this module still require `MonadUnliftIO`, and most of the provided instances in this PR can't support `MonadUnliftIO`.

For the ones that do support it, like `ReaderT`, there is a tradeoff with these instances you are proposing where using `ReaderT Tracer IO a` wouldn't be use the `Tracer` because it is lifted to the base of the stack.

I think I would rather deprecate this module since it doesn't really fit with the rest of the library's design.

kakkun61:
> The variants of `inSpan`, `inSpan'`, `inSpan''` in this module still require `MonadUnliftIO`, and most of the provided instances in this PR can't support `MonadUnliftIO`.

Oh I see! You are right.

> For the ones that do support it, like `ReaderT`, there is a tradeoff with these instances you are proposing where using `ReaderT Tracer IO a` wouldn't be use the `Tracer` because it is lifted to the base of the stack.

Yes, but users needs to define an orphan instance when this library does not serve `instance MonadTracer m => MonadTracer (ReaderT r m)`. And users can define `instance MonadTracer (ReaderT Tracer m)` on necessary (maybe an `OVERLOADS` pragma required).

And so I think that this library should expose the instance.

kakkun61:
@iand675 No chance to pull this?

iand675:
Sorry for the delay! Since the reader instance is overlappable I guess we can merge this.

iand675 (approved):

kakkun61:
Thanks for review.

PR #39, opened 2022-10-14
Add a vendor-honeycomb package with trace link stuff

lf-:
We've been using something like this at Mercury for a while, but it needed rewriting due to its reliance on Honeycomb classic.


vendors/honeycomb/src/OpenTelemetry/Vendor/Honeycomb.hs
```diff
@@ -0,0 +1,329 @@
...
+  let resources = getMaterializedResourcesAttributes . getTracerProviderResources $ tracer
+  pure $
+    HoneycombTarget team <$> case mEnvName of
+      -- There is an env name -> Current-Honeycomb
+      Just envName -> do
+        AttributeValue (TextAttribute serviceName) <- lookupAttribute resources "service.name"
+        pure $ Current envName (DatasetName serviceName)
+      -- Honeycomb Classic
+      Nothing -> do
+        pure $ Classic (Config.defaultDataset cfg)
```
iand675, line 169: Just checking, will it default to classic if an error occurs? I don't remember the code well enough on the `honeycomb` package side.

iand675 (approved):

vendors/honeycomb/src/OpenTelemetry/Vendor/Honeycomb.hs
lf-: I'm not entirely sure what you mean. If we are not hardcoding the team name, it's always required to make an HTTP request to Honeycomb to figure out where the data is going regardless of Classic or Current. In the Classic case, it is required to get the team name and in Current, we need both team name and environment (which we can't get away with hardcoding for even in proprietary software).

Classic is distinguished by the auth response giving an empty string in the environment field.

PR #42, opened 2022-10-28
Fix batch shutdown deadlock

### LINKED ISSUE #41: Async exception safety in the batch exporter
The following is a stream of consciousness diagnosis and debugging of a hanging shutdown: 

# Debugging the shutdownTracerProvider

I noticed that the test suite was failing to cleanly exit.
The final thing in the app is `withGlobalTracing`, which looks like:

```
bracket initializeTracing shutdownTraceProvider (const action)
```

This `bracket` comes from `UnliftIO.Exception`, which means that
`shutdownTraceProvider` runs in an `uninterruptibleMask`. This makes it pretty
fragile - if it blocks, it may become unkillable.

Turning it into a `Control.Exception.bracket`, which does not mask the cleanup
action, did allow the shutdown to complete. So something potentially odd is
going on here!

I did a crude timer on the shutdown, and it took 30 seconds.
That's a suspicious amount - seems like a timeout to me.

```haskell
-- | This method provides a way for provider to do any cleanup required.
--
-- This will also trigger shutdowns on all internal processors.
--
-- @since 0.0.1.0
shutdownTracerProvider :: MonadIO m => TracerProvider -> m ()
shutdownTracerProvider TracerProvider{..} = liftIO $ do
  asyncShutdownResults <- forM tracerProviderProcessors $ \processor -> do
    processorShutdown processor
  mapM_ wait asyncShutdownResults
```

So, we'll look at each processor, call `processShutdown` on it, and then `wait`
on those `async`s to finish. `processorShutdown` is a field on the `Processor`
record:

```haskell
data Processor = Processor
  { processorOnStart :: IORef ImmutableSpan -> Context -> IO ()
  -- ^ Called when a span is started. This method is called synchronously on the thread that started the span, therefore it should not block or throw exceptions.
  , processorOnEnd :: IORef ImmutableSpan -> IO ()
  -- ^ Called after a span is ended (i.e., the end timestamp is already set). This method is called synchronously within the 'OpenTelemetry.Trace.endSpan' API, therefore it should not block or throw an exception.
  , processorShutdown :: IO (Async ShutdownResult)
  -- ^ Shuts down the processor. Called when SDK is shut down. This is an opportunity for processor to do any cleanup required.
  --
  -- Shutdown SHOULD be called only once for each SpanProcessor instance. After the call to Shutdown, subsequent calls to OnStart, OnEnd, or ForceFlush are not allowed. SDKs SHOULD ignore these calls gracefully, if possible.
  --
  -- Shutdown SHOULD let the caller know whether it succeeded, failed or timed out.
  --
  -- Shutdown MUST include the effects of ForceFlush.
  --
  -- Shutdown SHOULD complete or abort within some timeout. Shutdown can be implemented as a blocking API or an asynchronous API which notifies the caller via a callback or an event. OpenTelemetry client authors can decide if they want to make the shutdown timeout configurable.
{- snip ... -}
```

As a record field, we now need to find where this provider is created.
Fortunately, that's right there in `bracket`:

```haskell
-- src/Mercury/Tracing.hs

initializeTracing :: MonadUnliftIO m => m TracerProvider
initializeTracing = do
  (processors, tracerOptions') <- liftIO getTracerProviderInitializationOptions

  -- FIXME: this is probably an upstream bug in hs-opentelemetry, but regardless
  -- Throw away all spans if the OTLP endpoint is not configured: anyone
  -- exporting to Honeycomb would have to have set this, so it should be fine
  otlpEndpoint <- liftIO $ lookupEnv "OTEL_EXPORTER_OTLP_ENDPOINT"
  let processors' = case otlpEndpoint of
        Just _ -> processors
        Nothing -> []

  provider <- createTracerProvider processors' tracerOptions'
  setGlobalTracerProvider provider
  pure provider
```

So we use the processors for the default, but only if the otlpEndpoint is
populated.

```haskell
-- src/OpenTelemetry/Trace.hs

getTracerProviderInitializationOptions :: IO ([Processor], TracerProviderOptions)
getTracerProviderInitializationOptions = getTracerProviderInitializationOptions' (mempty :: Resource 'Nothing)

getTracerProviderInitializationOptions' :: (ResourceMerge 'Nothing any ~ 'Nothing) => Resource any -> IO ([Processor], TracerProviderOptions)
getTracerProviderInitializationOptions' rs = do
  sampler <- detectSampler
  attrLimits <- detectAttributeLimits
  spanLimits <- detectSpanLimits
  propagators <- detectPropagators
  processorConf <- detectBatchProcessorConfig
  exporters <- detectExporters
  builtInRs <- detectBuiltInResources
  envVarRs <- mkResource . map Just <$> detectResourceAttributes
  let allRs = mergeResources (builtInRs <> envVarRs) rs
  processors <- case exporters of
    [] -> do
      pure []
    e:_ -> do
      pure <$> batchProcessor processorConf e
  let providerOpts = emptyTracerProviderOptions
        { tracerProviderOptionsIdGenerator = defaultIdGenerator
        , tracerProviderOptionsSampler = sampler
        , tracerProviderOptionsAttributeLimits = attrLimits
        , tracerProviderOptionsSpanLimits = spanLimits
        , tracerProviderOptionsPropagators = propagators
        , tracerProviderOptionsResources = materializeResources allRs
        }
  pure (processors, providerOpts)

-- src/OpenTelemetry/Trace/Core.hs

-- | Initialize a new tracer provider
--
-- You should generally use 'getGlobalTracerProvider' for most applications.
createTracerProvider :: MonadIO m => [Processor] -> TracerProviderOptions -> m TracerProvider
createTracerProvider ps opts = liftIO $ do
  let g = tracerProviderOptionsIdGenerator opts
  pure $ TracerProvider
    (V.fromList ps)
    g
    (tracerProviderOptionsSampler opts)
    (tracerProviderOptionsResources opts)
    (tracerProviderOptionsAttributeLimits opts)
    (tracerProviderOptionsSpanLimits opts)
    (tracerProviderOptionsPropagators opts)
    (tracerProviderOptionsLogger opts)
```

So `createTracerProvider` isn't doing anything fancy.

`processors` is either an empty list, or if we have at least one exporter, then
we create the `batchProcessor`. Nothing else modifies it, so we're good to just
look at the function:

```haskell
-- |
-- The batch processor accepts spans and places them into batches. Batching helps better compress the data and reduce the number of outgoing connections
-- required to transmit the data. This processor supports both size and time based batching.
--
batchProcessor :: MonadIO m => BatchTimeoutConfig -> Exporter ImmutableSpan -> m Processor
batchProcessor BatchTimeoutConfig{..} exporter = liftIO $ do
  batch <- newIORef $ boundedMap maxQueueSize
  workSignal <- newEmptyMVar
  worker <- async $ loop $ do
    req <- liftIO $ timeout (millisToMicros scheduledDelayMillis)
      $ takeMVar workSignal
    batchToProcess <- liftIO $ atomicModifyIORef' batch buildExport
    res <- liftIO $ Exporter.exporterExport exporter batchToProcess

    -- if we were asked to shutdown, quit cleanly after this batch
    -- FIXME: this could lose batches if there's more than one in queue?
    case req of
      Just Shutdown -> throwE res
      _ -> pure ()

  pure $ Processor
    { processorOnStart = \_ _ -> pure ()
    , processorOnEnd = \s -> do
        span_ <- readIORef s
        appendFailed <- atomicModifyIORef' batch $ \builder ->
          case push span_ builder of
            Nothing -> (builder, True)
            Just b' -> (b', False)
        when appendFailed $ void $ tryPutMVar workSignal Flush

    , processorForceFlush = void $ tryPutMVar workSignal Flush
    -- TODO where to call restore, if anywhere?
    , processorShutdown = async $ mask $ \_restore -> do
        -- flush remaining messages
        void $ tryPutMVar workSignal Shutdown

        shutdownResult <- timeout (millisToMicros exportTimeoutMillis) $
          wait worker
        -- make sure the worker comes down
        uninterruptibleCancel worker
        -- TODO, not convinced we should shut down processor here

        case shutdownResult of
          Nothing -> pure ShutdownFailure
          Just _ -> pure ShutdownSuccess
    }
  where
    millisToMicros = (* 1000)
```

There's a lot here, so let's just look at `processShutdown` - that's the actual
code we're running.

```haskell
    -- TODO where to call restore, if anywhere?
    , processorShutdown = async $ mask $ \_restore -> do
        -- flush remaining messages
        void $ tryPutMVar workSignal Shutdown

        shutdownResult <- timeout (millisToMicros exportTimeoutMillis) $
          wait worker
        -- make sure the worker comes down
        uninterruptibleCancel worker
        -- TODO, not convinced we should shut down processor here

        case shutdownResult of
          Nothing -> pure ShutdownFailure
          Just _ -> pure ShutdownSuccess
```

Well, we have a `mask` there, and we never `restore` it. But the original code
was in an `uninterruptibleMask` thanks to `bracket` anyway.

Line by line, let's look at these:

```haskell
        -- flush remaining messages
        void $ tryPutMVar workSignal Shutdown
```

[`tryPutMVar`](https://www.stackage.org/haddock/lts-19.30/base-4.15.1.0/Control-Concurrent-MVar.html#v:tryPutMVar)
is guaranteed not to block, so unlikely to be a problem here.
However, if the `workSignal` is full, then the `Shutdown` message is not
received.

```haskell
        shutdownResult <- timeout (millisToMicros exportTimeoutMillis) $
          wait worker
```

This may be our culprit.`timeout` comes from `System.Timeout` in the `base`
package. The docs have a hint:

```haskell
-- |Wrap an 'IO' computation to time out and return @Nothing@ in case no result
-- is available within @n@ microseconds (@1\/10^6@ seconds). In case a result
-- is available before the timeout expires, @Just a@ is returned. A negative
-- timeout interval means \"wait indefinitely\". When specifying long timeouts,
-- be careful not to exceed @maxBound :: Int@.
--
-- >>> timeout 1000000 (threadDelay 1000 *> pure "finished on time")
-- Just "finished on time"
--
-- >>> timeout 10000 (threadDelay 100000 *> pure "finished on time")
-- Nothing
--
-- The design of this combinator was guided by the objective that @timeout n f@
-- should behave exactly the same as @f@ as long as @f@ doesn't time out. This
-- means that @f@ has the same 'myThreadId' it would have without the timeout
-- wrapper. Any exceptions @f@ might throw cancel the timeout and propagate
-- further up. It also possible for @f@ to receive exceptions thrown to it by
-- another thread.
--
-- A tricky implementation detail is the question of how to abort an @IO@
-- computation. This combinator relies on asynchronous exceptions internally
-- (namely throwing the computation the 'Timeout' exception).  The technique
-- works very well for computations executing inside of the Haskell runtime
-- system, but it doesn't work at all for non-Haskell code.  Foreign function
-- calls, for example, cannot be timed out with this combinator simply because
-- an arbitrary C function cannot receive asynchronous exceptions. When
-- @timeout@ is used to wrap an FFI call that blocks, no timeout event can be
-- delivered until the FFI call returns, which pretty much negates the purpose
-- of the combinator. In practice, however, this limitation is less severe than
-- it may sound. Standard I\/O functions like 'System.IO.hGetBuf',
-- 'System.IO.hPutBuf', Network.Socket.accept, or 'System.IO.hWaitForInput'
-- appear to be blocking, but they really don't because the runtime system uses
-- scheduling mechanisms like @select(2)@ to perform asynchronous I\/O, so it
-- is possible to interrupt standard socket I\/O or file I\/O using this
-- combinator.
---
-- Note that 'timeout' cancels the computation by throwing it the 'Timeout'
-- exception. Consequently blanket exception handlers (e.g. catching
-- 'SomeException') within the computation will break the timeout behavior.
```

Ah, so, alas, it throws an asynchronous exception. But *we are in an
uninterruptible masked state*. So the exception is never delivered to our
timeout. That explains why `UnliftIO.Exception.bracket` causes the problem.

(also, that timeout? that's the 30 seconds we're waiting on --
`batchTimeoutConfig` has a default of 30,000 milliseconds to wait)

Woof. That's why our thread is *not* dying, but why isn't it, uh, dying
correctly?

Well, back to the shutdown code - what happens next?

```haskell
        -- make sure the worker comes down
        uninterruptibleCancel worker
        -- TODO, not convinced we should shut down processor here
```

Huh. Not entirely clear what goes on here!

```haskell
-- | Cancel an asynchronous action
--
-- This is a variant of `cancel`, but it is not interruptible.
{-# INLINE uninterruptibleCancel #-}
uninterruptibleCancel :: AsyncM a -> IO ()
uninterruptibleCancel = uninterruptibleMask_ . cancel
```

OK, so we're not saying "you MUST cancel, NOW", we're saying, "We're going to
cancel you, and the *cancellation* is uninterruptible." Let's look at `cancel`
itself:

```haskell
-- | Cancel an asynchronous action by throwing the @AsyncCancelled@
-- exception to it, and waiting for the `Async` thread to quit.
-- Has no effect if the 'Async' has already completed.
--
-- > cancel a = throwTo (asyncThreadId a) AsyncCancelled <* waitCatch a
--
-- Note that 'cancel' will not terminate until the thread the 'Async'
-- refers to has terminated. This means that 'cancel' will block for
-- as long said thread blocks when receiving an asynchronous exception.
--
-- For example, it could block if:
--
-- * It's executing a foreign call, and thus cannot receive the asynchronous
-- exception;
-- * It's executing some cleanup handler after having received the exception,
-- and the handler is blocking.
{-# INLINE cancel #-}
cancel :: Async a -> IO ()
cancel a@(Async t _) = throwTo t AsyncCancelled <* waitCatch a
```

OK, so `cancel` actually *waits for the `Async` to finish*! So if the `Async`
isn't responding to shutdown for whatever reason, then `cancel` won't return.

It seems to me that we should just be using `cancel`. That way, if we receive an
async exception in our shutdown, then we don't completely freak out.

```haskell
        case shutdownResult of
          Nothing -> pure ShutdownFailure
          Just _ -> pure ShutdownSuccess
```

OK, this is pure, nothing funny or fancy here.

# Investigating the `worker`

OK, so `timeout` is why we're *not dying*, but the real reason is that we're
doing something that takes forever inside of that timeout - trying to do a `wait
worker`. So let's look at `worker` and see what it's doing.

```haskell
  worker <- async $ loop $ do
    req <- liftIO $ timeout (millisToMicros scheduledDelayMillis)
      $ takeMVar workSignal
    batchToProcess <- liftIO $ atomicModifyIORef' batch buildExport
    res <- liftIO $ Exporter.exporterExport exporter batchToProcess

    -- if we were asked to shutdown, quit cleanly after this batch
    -- FIXME: this could lose batches if there's more than one in queue?
    case req of
      Just Shutdown -> throwE res
      _ -> pure ()
```

`worker` is an `Async ExportResult`. `loop` is a neat trick:

```haskell
-- | Exitable forever loop
loop :: Monad m => ExceptT e m a -> m e
loop = liftM (either id id) . runExceptT . forever
```

`throwE` terminates the loop.

So, when the `worker` is created, we use `async`. When we call
`withGlobalTracing`, we aren't in a masked state. And the initialization doesn't
appear to do any masking either. So this should be cancellable.

We have another `timeout` - this time, we're trying to `takeMVar` from the
`workSignal`. If `workSignal` is empty, this will block. So we `takeMVar` for
about ~5000ms by default - 5 seconds. There are a few things that put things in
the work signal: `Shutdown` causes the loop to exit.

`Expoerter.exporterExport` is a record field:

```haskell
data Exporter a = Exporter
  { exporterExport :: HashMap InstrumentationLibrary (Vector a) -> IO ExportResult
  , exporterShutdown :: IO ()
  }
```

That was provided in `getTracerProviderInitializationOptions'`:

```haskell
  exporters <- detectExporters
  {- snip -}
  processors <- case exporters of
    [] -> do
      pure []
    e:_ -> do
      pure <$> batchProcessor processorConf e
```

So, where does this come from?

```haskell
knownExporters :: [(T.Text, IO (Exporter ImmutableSpan))]
knownExporters =
  [ ("otlp", do
      otlpConfig <- loadExporterEnvironmentVariables
      otlpExporter otlpConfig
    )
  , ("jaeger", error "Jaeger exporter not implemented")
  , ("zipkin", error "Zipkin exporter not implemented")
  ]

-- TODO, support multiple exporters
detectExporters :: IO [Exporter ImmutableSpan]
detectExporters = do
  exportersInEnv <- fmap (T.splitOn "," . T.pack) <$> lookupEnv "OTEL_TRACES_EXPORTER"
  if exportersInEnv == Just ["none"]
    then pure []
    else do
      let envExporters = fromMaybe ["otlp"] exportersInEnv
          exportersAndRegistryEntry = map (\k -> maybe (Left k) Right $ lookup k knownExporters) envExporters
          (_notFound, exporterIntializers) = partitionEithers exportersAndRegistryEntry
      -- TODO, notFound logging
      sequence exporterIntializers
```

So, the only real option appears to be `otlpExporter`. Digging in... and it's a
big function, so skim it real quick.

```haskell
-- | Initial the OTLP 'Exporter'
otlpExporter :: (MonadIO m) => OTLPExporterConfig -> m (Exporter OT.ImmutableSpan)
otlpExporter conf = do
  -- TODO, url parsing is janky
-- TODO configurable retryDelay, maximum retry counts
  req <- liftIO $ parseRequest (maybe "http://localhost:4318/v1/traces" (<> "/v1/traces") (otlpEndpoint conf))

  let (encodingHeader, encoder) = maybe (id, id)
        (\case
          None -> (id, id)
          GZip -> (((hContentEncoding, "gzip") :), compress)
        )
        (otlpTracesCompression conf <|> otlpCompression conf)

      baseReqHeaders = encodingHeader $
        (hContentType, protobufMimeType) :
        (hAcceptEncoding, protobufMimeType) :
        fromMaybe [] (otlpHeaders conf) ++
        fromMaybe [] (otlpTracesHeaders conf) ++
        requestHeaders req
      baseReq = req
        { method = "POST"
        , requestHeaders = baseReqHeaders
        }
  pure $ Exporter
    { exporterExport = \spans_ -> do
        let anySpansToExport = H.size spans_ /= 0 && not (all V.null $ H.elems spans_)
        if anySpansToExport
          then do
            result <- try $ exporterExportCall encoder baseReq spans_
            case result of
              Left err -> do
                print err
                pure $ Failure $ Just err
              Right ok -> pure ok
          else pure Success
    , exporterShutdown = pure ()
    }
  where
    retryDelay = 100_000 -- 100ms
    maxRetryCount = 5
    isRetryableStatusCode status_ =
      status_ == status408 || status_ == status429 || (statusCode status_ >= 500 && statusCode status_ < 600)
    isRetryableException = \case
      ResponseTimeout -> True
      ConnectionTimeout -> True
      ConnectionFailure _ -> True
      ConnectionClosed -> True
      _ -> False

    exporterExportCall encoder baseReq spans_ = do
      msg <- encodeMessage <$> immutableSpansToProtobuf spans_
      -- TODO handle server disconnect
      let req = baseReq
            { requestBody =
                RequestBodyLBS $ encoder $ L.fromStrict msg
            }
      sendReq req 0 -- TODO =<< getTime for maximum cutoff

    sendReq req backoffCount = do
      eResp <- try $ httpBS req

      let exponentialBackoff = if backoffCount == maxRetryCount
            then pure $ Failure Nothing
            else do
              threadDelay (retryDelay `shiftL` backoffCount)
              sendReq req (backoffCount + 1)

      case eResp of
        Left err@(HttpExceptionRequest _ e) -> if isRetryableException e
          then exponentialBackoff
          else pure $ Failure $ Just $ SomeException err
        Left err -> pure $ Failure $ Just $ SomeException err
        Right resp -> if isRetryableStatusCode (responseStatus resp)
          then case lookup hRetryAfter $ responseHeaders resp of
            Nothing -> exponentialBackoff
            Just retryAfter -> do
              -- TODO support date in retry-after header
              case readMaybe $ C.unpack retryAfter of
                Nothing -> exponentialBackoff
                Just seconds -> do
                  threadDelay (seconds * 1_000_000)
                  sendReq req (backoffCount + 1)

          else pure $! if statusCode (responseStatus resp) >= 300
            then Failure Nothing
            else Success
```

Wow! OK, that's pretty huge. The active code appears to be the `then` branch
inside of the record field:

```haskell
        if anySpansToExport
          then do
            result <- try $ exporterExportCall encoder baseReq spans_
            case result of
              Left err -> do
                print err
                pure $ Failure $ Just err
              Right ok -> pure ok
          else pure Success
```

printing? in my dependencies? it's more likely than you think

anyway, we're definitely doing something that can take a while. Looks like our
app is potentially doing a lot of `threadDelay` to try and get these things in.

Well, that could explain the behavior we're seeing - if we're stuck in a
`exponentialBackoff` that initially waits 100ms and doubles every failure, we
may see that behavior.

Just out of curiosity, it does seem like... this should eventually end? So I'm
just going to let it run for maybe 20 minutes or so and see what happens.
Starting at about 3:15 my time...

and at 3:24, it's still goin'. Not good. Something is up.

```haskell
            result <- try $ exporterExportCall encoder baseReq spans_
            case result of
              Left err -> do
                print err
                pure $ Failure $ Just err
```

That `err` is a `SomeException`

```haskell
data ExportResult
  = Success
  | Failure (Maybe SomeException)
```

which means we could have caught an async exception, since this is plain
`Control.Exception`:

```haskell
import Control.Exception (SomeException(..), try)
```

But, if we got a `Failure`, then we should have handled that in the *call* to
the exporter, right? Let's jump back up the stack a bit, and see that exporter
call.

```haskell
  worker <- async $ loop $ do
    req <- liftIO $ timeout (millisToMicros scheduledDelayMillis)
      $ takeMVar workSignal
    batchToProcess <- liftIO $ atomicModifyIORef' batch buildExport
    res <- liftIO $ Exporter.exporterExport exporter batchToProcess

    -- if we were asked to shutdown, quit cleanly after this batch
    -- FIXME: this could lose batches if there's more than one in queue?
    case req of
      Just Shutdown -> throwE res
      _ -> pure ()
```

oops! We don't actually *do* anything with `res`! Success or failure, we just
kinda... ignore it.

So my new hypothesis is something like this:

The `exporterExport` caught an async exception in `try`, and this was dropped in
the `worker` loop. For some reason, we don't get the memo to shutdown, and the
timeout never completes, because the thread is blocked `wait`ing on the
`worker`.

So, what else can fill that `MVar`?

```haskell
    , processorOnEnd = \s -> do
        span_ <- readIORef s
        appendFailed <- atomicModifyIORef' batch $ \builder ->
          case push span_ builder of
            Nothing -> (builder, True)
            Just b' -> (b', False)
        when appendFailed $ void $ tryPutMVar workSignal Flush

    , processorForceFlush = void $ tryPutMVar workSignal Flush
```

These are our two candidates. They can do a `tryPutMVar` with `Flush`, which
would not shut the worker down.

What calls `processorOnEnd`? `endSpan`, which is used liberally in our codebase
as part of `inSpan = bracket createSpan endSpan`.

OK, let's run this through:

1. `worker` starts a loop iteration. The timeout starts.
2. `endSpan` happens, and we get a `Flush` in the `MVar`.
3. `worker` `takeMVar` succeeds, emptying the `MVar`.
4. `worker` then calls `exporterExport`, which begins the export. This may take
   *a long time* if the request fails for some reason!
5. `endSpan` happens, we get another `Flush` in the `MVar`.
6. BOOM, we've called `processorShutdown`. Now we `tryPutMVar Shutdown`, but
   this does *nothing* because it already has a `Flush`.
7. The `timeout` begins, `wait`ing for the worker to complete. But because the
   `Shutdown` write didn't take, it won't ever stop.
8. Then the `worker` gets a `cancel`, receiving an *async exception* to kill the
   thread. BUT... if this happens in that `try`? then we swallow the exception
   and keep going! Most likely, because that `try` surrounds a lot of
   `threadDelay`, we *will* throw to that `try`.

`wait worker` never completes.

---

I'm gonna work up a PR that should alleviate these issues.

parsonsmatt:
Fixes #41

### DIFF
--- a/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
+++ b/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
@@ -238,8 +238,7 @@ otlpExporter conf = do
                     Just (SomeAsyncException _) ->
                       throwIO err
                     Nothing ->
-                      pure ()
-                  pure $ Failure $ Just err
+                      pure $ Failure $ Just err
                 Right ok -> pure ok
             else pure Success
       , exporterShutdown = pure ()
--- a/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
+++ b/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
@@ -47,7 +47,7 @@ module OpenTelemetry.Exporter.OTLP (
 import Codec.Compression.GZip
 import Control.Applicative ((<|>))
 import Control.Concurrent (threadDelay)
-import Control.Exception (SomeException (..), SomeAsyncException(..), fromException, throwIO, try)
+import Control.Exception (SomeAsyncException(..), SomeException (..), fromException, throwIO, try)
 import Control.Monad.IO.Class
 import Data.Bits (shiftL)
 import qualified Data.ByteString.Char8 as C
--- a/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
+++ b/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
@@ -47,7 +47,7 @@ module OpenTelemetry.Exporter.OTLP (
 import Codec.Compression.GZip
 import Control.Applicative ((<|>))
 import Control.Concurrent (threadDelay)
-import Control.Exception (SomeAsyncException(..), SomeException (..), fromException, throwIO, try)
+import Control.Exception (SomeAsyncException (..), SomeException (..), fromException, throwIO, try)
 import Control.Monad.IO.Class
 import Data.Bits (shiftL)
 import qualified Data.ByteString.Char8 as C
--- a/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
+++ b/exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
@@ -47,7 +47,7 @@ module OpenTelemetry.Exporter.OTLP (
 import Codec.Compression.GZip
 import Control.Applicative ((<|>))
 import Control.Concurrent (threadDelay)
-import Control.Exception (SomeException (..), try)
+import Control.Exception (SomeException (..), SomeAsyncException(..), fromException, throwIO, try)
 import Control.Monad.IO.Class
 import Data.Bits (shiftL)
 import qualified Data.ByteString.Char8 as C
@@ -230,7 +230,15 @@ otlpExporter conf = do
               result <- try $ exporterExportCall encoder baseReq spans_
               case result of
                 Left err -> do
-                  print err
+                  -- If the exception is async, then we need to rethrow it
+                  -- here. Otherwise, there's a good chance that the
+                  -- calling code will swallow the exception and cause
+                  -- a problem.
+                  case fromException err of
+                    Just (SomeAsyncException _) ->
+                      throwIO err
+                    Nothing ->
+                      pure ()
                   pure $ Failure $ Just err
                 Right ok -> pure ok
             else pure Success
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -290,7 +290,7 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
               -- if a process is blocking.
 
               -- flush remaining messages and signal the worker to shutdown
-              void $ atomically $ putTMVar shutdownSignal Shutdown
+              void $ atomically $ putTMVar shutdownSignal ()
 
               -- gracefully wait for the worker to stop. we may be in
               -- a `bracket` or responding to an async exception, so we
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -22,8 +22,8 @@ module OpenTelemetry.Processor.Batch (
   -- , BatchProcessorOperations
 ) where
 
-import Control.Concurrent.STM
 import Control.Concurrent.Async
+import Control.Concurrent.STM
 import Control.Exception
 import Control.Monad
 import Control.Monad.IO.Class
@@ -195,7 +195,9 @@ buildExport m =
   , Builder.build <$> itemMap m
   )
 
+
 data ProcessorMessage = Flush | Shutdown
+
 
 -- note: [Unmasking Asyncs]
 --
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -309,11 +309,8 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
                         check shouldStop
                     ]
 
-              -- make sure the worker comes down. does not wait for the
-              -- worker thread to complete. wraps in SomeAsyncException to
-              -- ensure that any exception handler in the worker know sit
-              -- is an async exception.
-              throwTo (asyncThreadId worker) (SomeAsyncException AsyncCancelled)
+              -- make sure the worker comes down.
+              cancel worker
               -- TODO, not convinced we should shut down processor here
 
               pure $ case shutdownResult of
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -327,7 +327,6 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
   where
     millisToMicros = (* 1000)
 
-
 {-
 buffer <- newGreenBlueBuffer _ _
 batchProcessorAction <- async $ forever $ do
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -22,12 +22,11 @@ module OpenTelemetry.Processor.Batch (
   -- , BatchProcessorOperations
 ) where
 
+import Control.Concurrent.STM
 import Control.Concurrent.Async
-import Control.Concurrent.MVar (newEmptyMVar, takeMVar, tryPutMVar)
 import Control.Exception
 import Control.Monad
 import Control.Monad.IO.Class
-import Control.Monad.Trans.Except
 import Data.HashMap.Strict (HashMap)
 import qualified Data.HashMap.Strict as HashMap
 import Data.IORef (atomicModifyIORef', newIORef, readIORef)
@@ -36,7 +35,6 @@ import OpenTelemetry.Exporter (Exporter)
 import qualified OpenTelemetry.Exporter as Exporter
 import OpenTelemetry.Processor
 import OpenTelemetry.Trace.Core
-import System.Timeout
 import VectorBuilder.Builder as Builder
 import VectorBuilder.Vector as Builder
 
@@ -197,14 +195,30 @@ buildExport m =
   , Builder.build <$> itemMap m
   )
 
-
--- | Exitable forever loop
-loop :: Monad m => ExceptT e m a -> m e
-loop = liftM (either id id) . runExceptT . forever
-
-
 data ProcessorMessage = Flush | Shutdown
 
+-- note: [Unmasking Asyncs]
+--
+-- It is possible to create unkillable asyncs. Behold:
+--
+-- ```
+-- a <- uninterruptibleMask_ $ do
+--     async $ do
+--         forever $ do
+--             threadDelay 10_000
+--             putStrLn "still alive"
+-- cancel a
+-- ```
+--
+-- The prior code block will never successfully cancel `a` and will be
+-- blocked forever. The reason is that `cancel` sends an async exception to
+-- the thread performing the action, but the `uninterruptibleMask` state is
+-- inherited by the forked thread. This means that *no async exceptions*
+-- can reach it, and `cancel` will therefore run forever.
+--
+-- This also affects `timeout`, which uses an async exception to kill the
+-- running job. If the action is done in an uninterruptible masked state,
+-- then the timeout will not successfully kill the running action.
 
 {- |
  The batch processor accepts spans and places them into batches. Batching helps better compress the data and reduce the number of outgoing connections
@@ -213,20 +227,37 @@ data ProcessorMessage = Flush | Shutdown
 batchProcessor :: MonadIO m => BatchTimeoutConfig -> Exporter ImmutableSpan -> m Processor
 batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
   batch <- newIORef $ boundedMap maxQueueSize
-  workSignal <- newEmptyMVar
-  worker <- async $ loop $ do
-    req <-
-      liftIO $
-        timeout (millisToMicros scheduledDelayMillis) $
-          takeMVar workSignal
-    batchToProcess <- liftIO $ atomicModifyIORef' batch buildExport
-    res <- liftIO $ Exporter.exporterExport exporter batchToProcess
+  workSignal <- newEmptyTMVarIO
+  shutdownSignal <- newEmptyTMVarIO
+  let workerAction = do
+        delay <- registerDelay (millisToMicros scheduledDelayMillis)
+        req <- atomically $ do
+          msum
+            [ const Flush <$> do
+                continue <- readTVar delay
+                check continue
+            , const Flush <$> takeTMVar workSignal
+            , const Shutdown <$> takeTMVar shutdownSignal
+            ]
+        batchToProcess <- atomicModifyIORef' batch buildExport
+        -- we mask async exceptions in this, so that a buggy exporter that
+        -- catches async exceptions won't swallow them. since we use
+        -- an interruptible mask, blocking calls can still be killed, like
+        -- `threadDelay` or `putMVar` or most file I/O operations.
+        --
+        -- if we've received a shutdown, then we should be expecting
+        -- a `cancel` anytime now.
+        res <- mask_ $ Exporter.exporterExport exporter batchToProcess
 
-    -- if we were asked to shutdown, quit cleanly after this batch
-    -- FIXME: this could lose batches if there's more than one in queue?
-    case req of
-      Just Shutdown -> throwE res
-      _ -> pure ()
+        -- if we were asked to shutdown, quit cleanly after this batch
+        -- FIXME: this could lose batches if there's more than one in queue?
+        case req of
+          Shutdown ->
+            pure res
+          _ ->
+            workerAction
+  -- see note [Unmasking Asyncs]
+  worker <- asyncWithUnmask $ \unmask -> unmask workerAction
 
   pure $
     Processor
@@ -237,26 +268,65 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
             case push span_ builder of
               Nothing -> (builder, True)
               Just b' -> (b', False)
-          when appendFailed $ void $ tryPutMVar workSignal Flush
-      , processorForceFlush = void $ tryPutMVar workSignal Flush
+          when appendFailed $ void $ atomically $ tryPutTMVar workSignal ()
+      , processorForceFlush = void $ atomically $ tryPutTMVar workSignal ()
       , -- TODO where to call restore, if anywhere?
-        processorShutdown = async $ mask $ \_restore -> do
-          -- flush remaining messages
-          void $ tryPutMVar workSignal Shutdown
+        processorShutdown =
+          asyncWithUnmask $ \unmask -> unmask $ do
+            -- we call asyncWithUnmask here because the shutdown action is
+            -- likely to happen inside of a `finally` or `bracket`. the
+            -- @safe-exceptions@ pattern (followed by unliftio as well)
+            -- will use uninterruptibleMask in an exception cleanup. the
+            -- uninterruptibleMask state means that the `timeout` call
+            -- below will never exit, because `wait worker` will be in the
+            -- `uninterruptibleMasked` state, and the timeout async
+            -- exception will not be delivered.
+            --
+            -- see note [Unmasking Asyncs]
+            mask $ \_restore -> do
+              -- is it a little silly that we unmask and remask? seems
+              -- silly! but the `mask` here is doing an interruptible mask.
+              -- which means that async exceptions can still be delivered
+              -- if a process is blocking.
 
-          shutdownResult <-
-            timeout (millisToMicros exportTimeoutMillis) $
-              wait worker
-          -- make sure the worker comes down
-          uninterruptibleCancel worker
-          -- TODO, not convinced we should shut down processor here
+              -- flush remaining messages and signal the worker to shutdown
+              void $ atomically $ putTMVar shutdownSignal Shutdown
 
-          case shutdownResult of
-            Nothing -> pure ShutdownFailure
-            Just _ -> pure ShutdownSuccess
+              -- gracefully wait for the worker to stop. we may be in
+              -- a `bracket` or responding to an async exception, so we
+              -- must be very careful not to wait too long. the following
+              -- STM action will block, so we'll be susceptible to an async
+              -- exception.
+              delay <- registerDelay (millisToMicros exportTimeoutMillis)
+              shutdownResult <-
+                atomically $
+                  msum
+                    [ Just <$> waitCatchSTM worker
+                    , const Nothing <$> do
+                        shouldStop <- readTVar delay
+                        check shouldStop
+                    ]
+
+              -- make sure the worker comes down. does not wait for the
+              -- worker thread to complete. wraps in SomeAsyncException to
+              -- ensure that any exception handler in the worker know sit
+              -- is an async exception.
+              throwTo (asyncThreadId worker) (SomeAsyncException AsyncCancelled)
+              -- TODO, not convinced we should shut down processor here
+
+              pure $ case shutdownResult of
+                Nothing ->
+                  ShutdownTimeout
+                Just er ->
+                  case er of
+                    Left _ ->
+                      ShutdownFailure
+                    Right _ ->
+                      ShutdownSuccess
       }
   where
     millisToMicros = (* 1000)
+
 
 {-
 buffer <- newGreenBlueBuffer _ _
--- a/sdk/src/OpenTelemetry/Processor/Batch.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch.hs
@@ -231,17 +231,7 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
   batch <- newIORef $ boundedMap maxQueueSize
   workSignal <- newEmptyTMVarIO
   shutdownSignal <- newEmptyTMVarIO
-  let workerAction = do
-        delay <- registerDelay (millisToMicros scheduledDelayMillis)
-        req <- atomically $ do
-          msum
-            [ const Flush <$> do
-                continue <- readTVar delay
-                check continue
-            , const Flush <$> takeTMVar workSignal
-            , const Shutdown <$> takeTMVar shutdownSignal
-            ]
-        batchToProcess <- atomicModifyIORef' batch buildExport
+  let publish batchToProcess = mask_ $ do
         -- we mask async exceptions in this, so that a buggy exporter that
         -- catches async exceptions won't swallow them. since we use
         -- an interruptible mask, blocking calls can still be killed, like
@@ -249,13 +239,37 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
         --
         -- if we've received a shutdown, then we should be expecting
         -- a `cancel` anytime now.
-        res <- mask_ $ Exporter.exporterExport exporter batchToProcess
+        Exporter.exporterExport exporter batchToProcess
 
-        -- if we were asked to shutdown, quit cleanly after this batch
-        -- FIXME: this could lose batches if there's more than one in queue?
+  let flushQueueImmediately ret = do
+        batchToProcess <- atomicModifyIORef' batch buildExport
+        if null batchToProcess
+          then do
+            pure ret
+          else do
+            ret' <- publish batchToProcess
+            flushQueueImmediately ret'
+
+  let waiting = do
+        delay <- registerDelay (millisToMicros scheduledDelayMillis)
+        atomically $ do
+          msum
+            [ Flush <$ do
+                continue <- readTVar delay
+                check continue
+            , Flush <$ takeTMVar workSignal
+            , Shutdown <$ takeTMVar shutdownSignal
+            ]
+
+  let workerAction = do
+        req <- waiting
+        batchToProcess <- atomicModifyIORef' batch buildExport
+        res <- publish batchToProcess
+
+        -- if we were asked to shutdown, stop waiting and flush it all out
         case req of
           Shutdown ->
-            pure res
+            flushQueueImmediately res
           _ ->
             workerAction
   -- see note [Unmasking Asyncs]


sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -213,20 +227,37 @@ data ProcessorMessage = Flush | Shutdown
 batchProcessor :: MonadIO m => BatchTimeoutConfig -> Exporter ImmutableSpan -> m Processor
 batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
   batch <- newIORef $ boundedMap maxQueueSize
-  workSignal <- newEmptyMVar
-  worker <- async $ loop $ do
-    req <-
-      liftIO $
-        timeout (millisToMicros scheduledDelayMillis) $
-          takeMVar workSignal
```
parsonsmatt, line 221: Two possible problems with this code:

1. In an `uninterruptibleMask`, the `timeout` can never kill the `takeMVar`, and this could potentially never complete (until someone manually calls the flush or shutdown and the `MVar` is filled
2. It is possible for the `MVar` to be filled, and later calls to `tryPutMVar` fail. Suppose we had this flow:
    1. `MVar` is empty
    2. `takeMVar` blocks
    3. `tryPutMVar Flush` happens, `takeMVar` succeeds and proceeds to make HTTP calls
    4. `tryPutMVar Flush` happens, and succeeds
    5. `tryPutMVar Shutdown` happens, fails to write
    6. HTTP calls are done, `takeMVar` succeeds returning `Flush`.

Switching to two separate variables allows a separate "shutdown" and "flush now" flow. But using two variables means we want `STM`.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -213,20 +227,37 @@ data ProcessorMessage = Flush | Shutdown
...
+  let workerAction = do
+        delay <- registerDelay (millisToMicros scheduledDelayMillis)
+        req <- atomically $ do
+          msum
+            [ const Flush <$> do
+                continue <- readTVar delay
+                check continue
+            , const Flush <$> takeTMVar workSignal
+            , const Shutdown <$> takeTMVar shutdownSignal
+            ]
```
parsonsmatt, line 241: This is an async exception safe way to deal with `timeout` and reading from a variable. `registerDelay` forks a thread that merely writes to the `TVar Bool` when the elapsed time has completed. `takeTMVar` will block until the var is full, and then empty it.

Suppose we have a write to `Flush` and `Shutdown` at the same time. If the RTS picks `Flush`, it's no big deal - we run the HTTP action, then we loop again. The `Shutdown` var will be full, and will "flush" and then shutdown.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -213,20 +227,37 @@ data ProcessorMessage = Flush | Shutdown
...
+        batchToProcess <- atomicModifyIORef' batch buildExport
+        -- we mask async exceptions in this, so that a buggy exporter that
+        -- catches async exceptions won't swallow them. since we use
+        -- an interruptible mask, blocking calls can still be killed, like
+        -- `threadDelay` or `putMVar` or most file I/O operations.
+        --
+        -- if we've received a shutdown, then we should be expecting
+        -- a `cancel` anytime now.
+        res <- mask_ $ Exporter.exporterExport exporter batchToProcess
```
parsonsmatt, line 250: i'm not 100% happy with this, but the ability of any `Exporter` being able to `try x :: IO (Either SomeException a)` and catch async exceptions makes this pretty tricky.

Realistically, any exporter is going to be doing `threadDelay` or some other kind of blocking IO, and it'll receive the async exception anyway. `mask_` or not probably doesn't matter. Eg: `try $ threadDelay 10000` will still catch the async exception.

If we wanted to be *really* sure about the action not swallowing async exceptions, we'd need to put it in an`uninterruptibleMask`. *IF* the action completes, then the async exception would be received after the `uninterruptibleMask` exits. But there's no guarantee that the action wouldn't block or loop infinitely, so that seems pretty dicey.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,27 +268,66 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+            --
+            -- see note [Unmasking Asyncs]
+            mask $ \_restore -> do
+              -- is it a little silly that we unmask and remask? seems
+              -- silly! but the `mask` here is doing an interruptible mask.
+              -- which means that async exceptions can still be delivered
+              -- if a process is blocking.
+
+              -- flush remaining messages and signal the worker to shutdown
+              void $ atomically $ putTMVar shutdownSignal ()
```
parsonsmatt, line 293: attempt at graceful shutdown #1...

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,27 +268,66 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+              -- exception.
+              delay <- registerDelay (millisToMicros exportTimeoutMillis)
+              shutdownResult <-
+                atomically $
+                  msum
+                    [ Just <$> waitCatchSTM worker
+                    , const Nothing <$> do
+                        shouldStop <- readTVar delay
+                        check shouldStop
+                    ]
```
parsonsmatt, line 308: wait for the worker to terminate, but only wait for the expected delay time

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,27 +268,66 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+
+              -- make sure the worker comes down. does not wait for the
+              -- worker thread to complete. wraps in SomeAsyncException to
+              -- ensure that any exception handler in the worker know sit
+              -- is an async exception.
+              throwTo (asyncThreadId worker) (SomeAsyncException AsyncCancelled)
```
parsonsmatt, line 314: Previously this was `uninterruptibleCancel` - `uninterruptibleMask_ . cancel`. And `cancel (Async tid wait) = atomically wait <* throwTo tid AsyncCancelled`. So.... `cancel` actually *blocks* until the async thread has terminated.

With `uninterruptibleCancel`, if the `Async` is *stuck*, then we're also *stuck*, and the only way to escape is with the RTS killing everything abruptly.

Rather than kill everything abruptly, let's just *not block forever*, and hope that the worker isn't irreparably stuck.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,27 +268,66 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+
+              pure $ case shutdownResult of
+                Nothing ->
+                  ShutdownTimeout
+                Just er ->
+                  case er of
+                    Left _ ->
+                      ShutdownFailure
+                    Right _ ->
+                      ShutdownSuccess
```
parsonsmatt, line 325: Previously the logic conflated `ShutdownTimeout` and `ShutdownFailure`. Now we can distinguish them.

parsonsmatt (commented):
I've verified that this fixes the deadlock locally.

parsonsmatt:
Uhhh - I ran `make format` locally in a `nix-shell` and it totally reformatted everything. I'm *guessing* that's not what we want, so I'm going to leave this up here.

sdk/src/OpenTelemetry/Processor/Batch.hs
iand675: I think you're on the right track here, but it'd probably be good to add a note to the public `Exporter` interface cautioning about these instances.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -198,35 +196,70 @@ buildExport m =
...
+        -- catches async exceptions won't swallow them. since we use
+        -- an interruptible mask, blocking calls can still be killed, like
+        -- `threadDelay` or `putMVar` or most file I/O operations.
+        --
+        -- if we've received a shutdown, then we should be expecting
+        -- a `cancel` anytime now.
+        res <- mask_ $ Exporter.exporterExport exporter batchToProcess
+
+        -- if we were asked to shutdown, quit cleanly after this batch
+        -- FIXME: this could lose batches if there's more than one in queue?
```
iand675, line 255: I'd really love to see this addressed before merging.

iand675 (commented):
Substantive improvements here, thanks for these! I'd love to see a regression test as part of this PR.

iand675 (commented):
Substantive improvements here, thanks for these! I'd love to see a regression test as part of this PR.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -198,35 +196,70 @@ buildExport m =
...
-      _ -> pure ()
+  workSignal <- newEmptyTMVarIO
+  shutdownSignal <- newEmptyTMVarIO
+  let workerAction = do
+        delay <- registerDelay (millisToMicros scheduledDelayMillis)
+        req <- atomically $ do
+          msum
+            [ const Flush <$> do
+                continue <- readTVar delay
+                check continue
```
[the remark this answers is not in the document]
parsonsmatt: `STM`'s `<|>` is defined as [`orElse`](https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Conc-Sync.html#v:orElse), and `msum` is `asum` is `foldr (<|>) empty`. 

`orElse` haddock:

> Compose two alternative STM actions (GHC only).
>
> If the first action completes without retrying then it forms the result of the [orElse](https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Conc-Sync.html#v:orElse). Otherwise, if the first action retries, then the second action is tried in its place. If both actions retry then the [orElse](https://hackage.haskell.org/package/base-4.17.0.0/docs/GHC-Conc-Sync.html#v:orElse) as a whole retries.

sdk/src/OpenTelemetry/Processor/Batch.hs
parsonsmatt: Hmm. I think we need to be able to distinguish between normal program exit and exceptional exit.

If we reinterpret `Shutdown` as "flush the queue until empty," instead of "shut down after the next flush," then we can have the following behavior:

On normal program exit, we want to block until the queue is flushed and everything is reported. Then this cleanup is done and we can go on to the next cleanup. Otherwise we lose metrics. We're not in a hurry, since we're exiting "normally."

But on exceptional exit, we *don't* want to block - we want to GTFO ASAP, so that we can run any other finalizers. In this case, we'd probably want to send a `Shutdown` to the worker thread, and exit immediately instead of `wait` or `cancel` or in any other way killing the thread. We should expect that the RTS will be reapin' us soon anyway, so we don't need to worry about anything too Weird happening.

If we cannot distinguish exceptional vs unexceptional, then we must accept the same behavior for both. In the "happy" case, the "exception-safe" behavior will allow the program to terminate as soon as the last finalizer is complete - potentially losing traces. In the "exception" path, the "happy" behavior will potentially block other resources from being cleaned up.

But allowing for exception vs unexceptional shutdown would require re-working the API a good bit more, I think.

iand675: OK, I'm convinced it's probably not necessary for this PR.

That said, I am fairly convinced that we still want to fully process the queue on exceptional exit– if we lose sampled traces that show *why* a program is exiting exceptionally, it reduces the utility of the library for short-lived processes quite a bit . I would argue that since consumers of the library can specify max queue size, batch export size, and export timeout, they get to control the maximum amount of shutdown time required.

iand675: Here's what the specification says about shutdown, FWIW: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown-1

parsonsmatt: > That said, I am fairly convinced that we still want to fully process the queue on exceptional exit– if we lose sampled traces that show why a program is exiting exceptionally, it reduces the utility of the library for short-lived processes quite a bit 

That makes sense to me. I suppose the *most* common case is that you start this up right near the very top of `main`, so there wouldn't be any other cleanup actions to run.

exporters/otlp/src/OpenTelemetry/Exporter/OTLP.hs
```diff
@@ -230,7 +230,15 @@ otlpExporter conf = do
...
-                  print err
+                  -- If the exception is async, then we need to rethrow it
+                  -- here. Otherwise, there's a good chance that the
+                  -- calling code will swallow the exception and cause
+                  -- a problem.
+                  case fromException err of
+                    Just (SomeAsyncException _) ->
+                      throwIO err
+                    Nothing ->
+                      pure ()
```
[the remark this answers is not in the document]
parsonsmatt: Good call!

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,23 +284,58 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+            -- below will never exit, because `wait worker` will be in the
+            -- `uninterruptibleMasked` state, and the timeout async
+            -- exception will not be delivered.
+            --
+            -- see note [Unmasking Asyncs]
+            mask $ \_restore -> do
+              -- is it a little silly that we unmask and remask? seems
+              -- silly! but the `mask` here is doing an interruptible mask.
+              -- which means that async exceptions can still be delivered
+              -- if a process is blocking.
```
lf-, line 306: question: given that we `unmask` here, there would be a race condition where someone could *theoretically* stuff in an inconveniently placed exception between the `unmask` and `mask` and stop our clean shutdown here, no?

I feel like we want to transition directly from `MaskedUninterruptible` -> `MaskedInterruptible`, rather than `MaskedUninterruptible` -> `Unmasked` -> `MaskedInterruptible`. https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Exception.html#t:MaskingState

Reading the docs of Control.Exception, I think that we want to use the function `interruptible` instead?

God, this is way too hard code; I really wish that this were abstracted into a library or something so it only has to be right once.

sdk/src/OpenTelemetry/Processor/Batch.hs
```diff
@@ -237,23 +284,58 @@ batchProcessor BatchTimeoutConfig {..} exporter = liftIO $ do
...
+              -- gracefully wait for the worker to stop. we may be in
+              -- a `bracket` or responding to an async exception, so we
+              -- must be very careful not to wait too long. the following
+              -- STM action will block, so we'll be susceptible to an async
+              -- exception.
+              delay <- registerDelay (millisToMicros exportTimeoutMillis)
+              shutdownResult <-
+                atomically $
+                  msum
+                    [ Just <$> waitCatchSTM worker
```
lf-, line 320: I don't quite follow how `waitCatchSTM` does not block here, but STM blocking semantics seem overall quite very nonobvious to me. My understanding here is that `waitCatchSTM` will `retry` if it not done, then this code will run the check for the timer. Is that correct?

lf- (commented):
I have left some comments. I don't understand this enough to be confident it works for sure.

sdk/src/OpenTelemetry/Processor/Batch.hs
parsonsmatt: You're right - we have a brief period of vulnerability to async exceptions.

```haskell
uninteruptibleMask_ $ 
  -- invincible
  asyncWithUnmask $ \unmask -> do
    -- still invincible
    unmask $ do
      -- totally vincible
      mask $ \restore -> do
        -- async exceptions only happen in blocking operations   
```

Fortunately, this period of vulnerability is very short, and I think it may not even be a problem. Async exceptions are only *really* visible at "yield points" (not a technical term) - with allocation being a yield point. `unsafeUnmask`, the [function that `unmask` here refers to](https://hackage.haskell.org/package/base-4.17.0.0/docs/src/GHC.IO.html#unsafeUnmask), unwraps the `IO` newtype and calls a primitive on it. `mask` might allocate if `getMaskingState` allocates, but I'd suspect that the allocation can be erased since we're immediately pattern matching on the result.

So uhh the safety here is *kinda* dependent on optimizations? Oof. Not sure we can really improve the situation. Especially since the only way we'll get this if we do somethin like

```haskell
cancel =<< shutdownProcessor processor
```

which, well, what do you expect from that?

`interruptible` is *almost* right: 

> When called outside [mask](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Exception.html#v:mask), or inside [uninterruptibleMask](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Exception.html#v:uninterruptibleMask), this function has no effect.

The problem is that we specifically need it if we're in `uninterruptibleMask`. It does seem like we're missing a primitive that does the transform you're talking about.

parsonsmatt: Follow up GHC issue: https://gitlab.haskell.org/ghc/ghc/-/issues/22389

sdk/src/OpenTelemetry/Processor/Batch.hs
parsonsmatt: `atomically` blocks if all of the actions `retry`. So this operation *does* block, and *can* throw an async exception, even inside of `mask`.

But, yeah, the logic is "First check `waitCatchSTM worker`. If that is not complete, then check the timer."

The GHC RTS is smart enough to not endlessly loop and continually retry, and will instead send a wakeup to any blocked `STM` action when any of the linked variables change. So `worker` here becomes a "watched" variable, as does `delay`. When `delay` gets updated by the other thread, then GHC will wake up this thread up, retry the transaction as a whole, and complete.

Now, it's possible we'd have this flow:

1. `atomically` starts and then blocks.
2. The timer thread finishes, and the `TVar Bool` becomes `True`.
3. *Before GHC can wake up the `atomically` thread*, the `waitCatchSTM` *also* finishes.
4. GHC wakes up `atomically`, which does `msum ...`.
5. `waitCatchSTM` is first, so it gets pulled.

But that's actually totally fine and what we want.

lf- (approved):
I think I've stared at this enough to believe it is correct. I'm fine with shipping this as is although regression tests would be nice.

parsonsmatt:
A difficulty with a proper regression test is that "test failure" would look like "the test suite hangs." Which, I guess we *could* do as a separate binary - one that triggers `processorFlush p >> processorShutdown p` one-after-the-other ought to do it, while in an `uninterruptibleMask`. Then we can `kill` the binary using a signal - killing = test failure, successful exit = test pass.

Which is a lot of infra for a fix - would y'all be satisfied with a follow-up issue? Maybe a "deadlock regression test suite" that can be used to autotest that a given `Processor` is resilient?

lf-:
I'm happy with the idea of filing a separate issue.

iand675:
Right, let's merge this then.

PR #46, opened 2022-12-15
sdk: Crash if -threaded is unset

lf-:
registerDelay will throw an exception, which we will silently swallow, if we
don't preemptively just blow up the program. It's never correct to start up
without a threaded runtime.


lf- (commented):
Note that rtsSupportsBoundThreads is the condition used by `registerDelay` (for some reason), which is why we use it.

parsonsmatt (approved):

iand675:
Seems reasonable. The spec states:

```
The API or SDK MAY fail fast and cause the application to fail on initialization, e.g. because of a bad user config or environment, but MUST NOT cause the application to fail later at run time, e.g. due to dynamic config settings received from the Collector
```

iand675:
That said, I think we should add a docstring mentioning this behavior.

iand675 (approved):

PR #50, opened 2023-01-17
persistent-mysql instrumentation

kakkun61:
- new instrumentation for persistent-mysql
- add `wrapSqlBackend'` to `OpenTelemetry.Instrumentation.Persistent`


instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
```diff
@@ -0,0 +1,110 @@
...
+createMySQLPool ::
+  (MonadUnliftIO m, MonadLoggerIO m) =>
+  Otel.TracerProvider ->
+  -- | Attributes that are specific to providers like MySQL, PostgreSQL, etc.
+  [(Text, Otel.Attribute)] ->
+  -- | Connection information.
+  MySQL.ConnectInfo ->
+  -- | Number of connections to be kept open in the pool.
+  Int ->
+  m (Pool SqlBackend)
```
iand675, line 52: I know that `hs-opentelemetry-instrumentation-persistent` currently only supports `SqlBackend`, but I wonder if we could add support for `IsSqlBackend backend` as part of this PR?

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
```diff
@@ -0,0 +1,110 @@
...
+  pure (conn, backend')
+
+
+{- | Same as 'withMySQLPool', but instead of opening a pool
+of connections, only one connection is opened.
+-}
+withMySQLConn ::
+  (MonadUnliftIO m, MonadLoggerIO m) =>
+  Otel.TracerProvider ->
+  -- | Attributes that are specific to providers like MySQL, PostgreSQL, etc.
```
iand675, line 98: I don't really see where any attributes are being added for MySQL connections in this PR. Is this expected to be a manual addition by applications?

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
kakkun61: How should I change after all?

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
kakkun61: No extra attributes are added by hs-opentelemetry-instrumentation-persistent-mysql now.

Ah, I found this document now. https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/database/

kakkun61:
That CI failure is caused by https://github.com/haskell/actions/issues/187.

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
kakkun61: I made some attributes be added at https://github.com/iand675/hs-opentelemetry/pull/50/commits/3fcbf3e863c8bba8aef2edd545261d9e5fa28a0a.

kakkun61:
marge #53 first

kakkun61:
@iand675 How about this PR?

iand675 (approved):

PR #54, opened 2023-03-01
dotenv for examples

kakkun61:
Add env vars for examples


.github/workflows/haskell.yml
```diff
@@ -12,6 +12,9 @@ jobs:
       - uses: haskell/actions/setup@v2
         with:
           enable-stack: true
+      - if: ${{ runner.os == 'Linux' }}
```
iand675, line 15: Are we able to drop the changes to this file?

.env
```diff
@@ -1 +1,2 @@
-PATH="./scripts:$PATH"
+# examples
+OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318
```
iand675, line 2: I'd rather that this `.env` file was scoped to the examples that use it. I think in the future we'll want to have examples for exporting to different 3rd party providers like datadog, so we'd need different env var values in those cases.

.envrc
```diff
@@ -1,3 +1,4 @@
 use_nix
+PATH_add "./scripts"
```
iand675, line 2: Nice!

iand675 (changes requested):

.github/workflows/haskell.yml
kakkun61: https://github.com/actions/runner-images/issues/7061 seems already been solved. Right!

.env
kakkun61: On such cases, we will have `OTEL_EXPORTER_XXX_ENDPOINT`s additionally?

Does `OTEL_EXPORTER_OTLP_ENDPOINT` existence make library use otlp protocol?

iand675: I think for the different examples, we would generally need different OTEL_EXPORTER_OTLP_ENDPOINT values. I don't think there are other standardized `OTEL_EXPORTER_XXX_ENDPOINT` values currently.

kakkun61: Okey 0a92377b1c863fcc3be4c1ee0601a16bfe4703b9

PR #56, opened 2023-03-01
example http server

kakkun61:
This example uses:

- hs-opentelemetry-api
- hs-opentelemetry-sdk
- hs-opentelemetry-propagator-datadog
- hs-opentelemetry-propagator-w3c
- hs-opentelemetry-instrumentation-wai
- hs-opentelemetry-instrumentation-http-client


examples/http-server/main.hs
```diff
@@ -0,0 +1,54 @@
+{-# LANGUAGE OverloadedStrings #-}
```
iand675, line 1: Small nitpick: `main.hs` should probably be `Main.hs` instead.

iand675 (changes requested):

kakkun61:
@iand675 I resolved this conflict.

PR #66, opened 2023-04-05
Adds B3 Propagator Library

solomon-b:
Implements the B3 Propagator library. 

Notes: 

- According to the [OpenTelemetry Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#b3-requirements) we do not propagate the B3 parent span id. 

- We are also unable to handle B3 single header deny decisions:
```
b3: 0
```
This is because we the `TraceContext` must contain a `TraceId` and `SpanId`.

- I did not build out the test suite but can add unit tests if desired.

### DIFF
--- a/propagators/b3/hs-opentelemetry-propagator-b3.cabal
+++ b/propagators/b3/hs-opentelemetry-propagator-b3.cabal
@@ -6,7 +6,9 @@ cabal-version: 1.12
 
 name:           hs-opentelemetry-propagator-b3
 version:        0.0.1.0
+synopsis:       Trace propagation via HTTP headers following the b3 tracestate spec.
 description:    Please see the README on GitHub at <https://github.com/iand675/hs-opentelemetry/tree/main/propagators/b3#readme>
+category:       OpenTelemetry, Tracing, Web
 homepage:       https://github.com/iand675/hs-opentelemetry#readme
 bug-reports:    https://github.com/iand675/hs-opentelemetry/issues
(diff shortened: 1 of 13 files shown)


solomon-b:
I've updated my branch to address the CI issues.

propagators/b3/hs-opentelemetry-propagator-b3.cabal
```diff
@@ -24,12 +26,23 @@ source-repository head
...
       Paths_hs_opentelemetry_propagator_b3
   hs-source-dirs:
       src
+  ghc-options: -Wall
   build-depends:
-      base >=4.7 && <5
+      attoparsec
+    , base >=4.7 && <5
+    , bytestring
+    , hs-opentelemetry-api ==0.0.3.*
```
kakkun61, line 41: Please update this _cabal_ file with the current _package.yaml_. `case-insensitive` is not included.

kakkun61 (changes requested):

propagators/b3/hs-opentelemetry-propagator-b3.cabal
solomon-b: Oh woops, we don't actually need `case-insensitive` and I removed it from the cabal file directly rather then `package.yaml`. I don't usually use hpack.

kakkun61 (approved):

PR #64, opened 2023-03-27
Fix memory leak

9999years:
Authored by: Ian-Woo Kim @wavewave

PR #68, opened 2023-04-21
refresh CI cache

kakkun61:
Some CI failed because of dirty CI cache (maybe).

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,9 +20,8 @@ jobs:
           path: |
             ~/.stack
             .stack-work
-          key: stack-${{ runner.os }}-8.12-${{ hashFiles('stack-ghc-8.12.yaml.lock') }}
+          key: stack-${{ runner.os }}-8.12-${{ hashFiles('stack-ghc-8.12.yaml.lock') }}-2
           restore-keys: |
-            stack-${{ runner.os }}-8.12
       - run: stack --stack-yaml stack-ghc-8.12.yaml build --only-dependencies
       - run: stack --stack-yaml stack-ghc-8.12.yaml build
       - run: stack --stack-yaml stack-ghc-8.12.yaml test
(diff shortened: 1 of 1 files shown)


kakkun61:
I think that this will fix CI failures of #56, #64 (some of), and #67.

lf-:
I wonder if I, as someone with write access, can just delete the cache with the api then rerun the actions. https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28

.github/workflows/haskell.yml
```diff
@@ -45,9 +44,8 @@ jobs:
           path: |
             ~/.stack
             .stack-work
-          key: stack-${{ runner.os }}-${{ hashFiles('stack.yaml.lock') }}
+          key: stack-${{ runner.os }}-${{ hashFiles('stack.yaml.lock') }}-2
           restore-keys: |
-            stack-${{ runner.os }}
```
lf-, line 50: Does this not disable caching altogether?

lf-:
I have run:
```
gh api /repos/iand675/hs-opentelemetry/actions/caches -q '.actions_caches | .[] | select(.ref == ("refs/pull/56/merge", "refs/pull/64/merge", "refs/pull/67/merge")) | .id' | xargs  -I '{}' gh api --method DELETE '/repos/iand675/hs-opentelemetry/actions/caches/{}'
```

.github/workflows/haskell.yml
kakkun61: When there is a cache whose key matches exactly, that cache is used.

kakkun61:
Wow nice api are there. If it runs expectedly, this PR is unnecessary.

.github/workflows/haskell.yml
lf-: Ah, I think this part of the PR should most likely be kept then, since I am not sure what clearing the caches might achieve if it continues to be broken. I am also slightly surprised/concerned that the caches are causing build breakage: Stack should be able to deal with this stuff properly!

PR #83, opened 2023-07-15
Add missing import

chrismanning:
Fixes build on ghc-9.6 at least

### DIFF
--- a/api/src/OpenTelemetry/Internal/Trace/Id.hs
+++ b/api/src/OpenTelemetry/Internal/Trace/Id.hs
@@ -52,7 +52,9 @@ import GHC.Exts (
   int2Word#,
   isTrue#,
   or#,
+#if MIN_VERSION_base(4,17,0)
   word64ToWord#,
+#endif
  )
 import GHC.Generics (Generic)
 import OpenTelemetry.Trace.Id.Generator (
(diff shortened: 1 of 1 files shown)


api/src/OpenTelemetry/Internal/Trace/Id.hs
```diff
@@ -52,6 +52,7 @@ import GHC.Exts (
   int2Word#,
   isTrue#,
   or#,
+  word64ToWord#,
```
kakkun61, line 55: It seems that base older than 4.17.0.0 does not expose `word64Word#`.

Please use CPP conditional branch.

kakkun61 (changes requested):

api/src/OpenTelemetry/Internal/Trace/Id.hs
```diff
@@ -52,6 +52,9 @@ import GHC.Exts (
   int2Word#,
   isTrue#,
   or#,
+#if MIN_VERSION_base(4,17,0)
+  word64ToWord#,
+#endif
```
kakkun61, line 57: Sadly fourmolu cannot parse this code, please write like this:

```haskell
#if MIN_VERSION_base(4,17,0)
import GHC.Exts (word64ToWord#)
#endif
```

kakkun61 (changes requested):

kakkun61 (approved):

PR #85, opened 2023-07-18
Implement max export batch size

jkeuhlen:
This PR adds support for `maxExportBatchSize` as described in [the spec](https://opentelemetry.io/docs/specs/otel/trace/sdk/#batching-processor): 
image


kakkun61:
Is it correct that this PR will merge into _improved-dux_ branch, not _main_ branch?

jkeuhlen:
@kakkun61 I based this branch off `improve-dux` since that's the one we track against internally right now (and thus where I started from to find this issue). I think this could easily be applied to main as well though. I'm not sure what the difference between the branches really is, or when one will be merged back into the other.

kakkun61:
I see!

PR #88, opened 2023-08-01
Use maps instead of lists of pairs as attributes

kakkun61:
Use `HashMap Text Attribute` instead of `[(Text, Attribute)]` as attributes because of more efficiency and more explicit constraints (for example about overlap of keys).

This is breaking change. So some packages' major versions must be bumped.


iand675 (approved):
Thank you for this!

PR #101, opened 2023-10-06
Release datadog propagator

kakkun61:


iand675:
@kakkun61 can you update this against main if it's still relevant?

PR #89, opened 2023-09-08
Swallow localhost errors

parsonsmatt:
By default, the program sends trace information to `localhost:4318`. However, if there isn't a collector running, you'll get spammed with error information. Consider this output:

```
hotel-california> Test suite hotel-california-test passed
Completed 2 action(s).        
("exit code: ",ExitSuccess)
HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 4318
  secure               = False
  requestHeaders       = [("Content-Type","application/x-protobuf"),("Accept-Encoding","application/x-protobuf")]
  path                 = "/v1/traces"
  queryString          = ""
  method               = "POST"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 87>: does not exist (Connection refused))
HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 4318
  secure               = False
  requestHeaders       = [("Content-Type","application/x-protobuf"),("Accept-Encoding","application/x-protobuf")]
  path                 = "/v1/traces"
  queryString          = ""
  method               = "POST"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 87>: does not exist (Connection refused))
HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 4318
  secure               = False
  requestHeaders       = [("Content-Type","application/x-protobuf"),("Accept-Encoding","application/x-protobuf")]
  path                 = "/v1/traces"
  queryString          = ""
  method               = "POST"
  proxy                = Nothing
  rawBody              = False
  redirectCount        = 10
  responseTimeout      = ResponseTimeoutDefault
  requestVersion       = HTTP/1.1
  proxySecureMode      = ProxySecureWithConnect
}
 (ConnectionFailure Network.Socket.connect: <socket: 87>: does not exist (Connection refused))
HttpExceptionRequest Request {
  host                 = "localhost"
  port                 = 4318
  secure               = False
  requestHeaders       = [("Content-Type","application/x-protobuf"),("Accept-Encoding","application/x-protobuf")]
[...]

PR #112, opened 2024-03-13
Loosen the bounds `sdk` has on `exporter-otlp`, so v0.0.2 (containing bugfixes of the last two years) can actually be used

Qqwy:
When using the libraries by depending on in `hs-opentelemetry-sdk`, you'll end up with a very outdated version of `hs-opentelemetry-exporter-otlp`, as the current bound was still `==0.0.1.*`. However version `0.0.2.0` contains many important bugfixes like for instance #42 and #89.

This PR loosens this bound.

### DIFF
--- a/exporters/otlp/hs-opentelemetry-exporter-otlp.cabal
+++ b/exporters/otlp/hs-opentelemetry-exporter-otlp.cabal
@@ -5,7 +5,7 @@ cabal-version: 1.12
 -- see: https://github.com/sol/hpack
 
 name:               hs-opentelemetry-exporter-otlp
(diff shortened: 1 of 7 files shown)


Qqwy:
@iand675 What do you want me to do with the failing format spec? It is complaining about a lot (all?) files, none of which I touched as I obviously only touched the `.yaml` and `.cabal` file in `./sdk`.

Maybe the version of `fourmolu` that is loaded inside the Github action was recently bumped?

Qqwy:
I also bumped the actual versions set in the package configs themselves as curiously some of them were behind their changelogs.

PR #116, opened 2024-05-07
Bracket WAI middleware spans with detachContext

velveteer:
After updating to `hs-opentelemetry-instrumentation-wai-0.1.0.0`, I noticed our Datadog APM traces were starting to group unrelated requests together. Notably all of these sporadic groups shared the same GHC thread ID.

Further investigation led me to this commit: https://github.com/iand675/hs-opentelemetry/commit/34569fc8a92189011cadeda757375847efc92ef7. At first glance, removing this `bracket` would seem harmless, assuming each WAI request is served by a separate thread. But in fact `warp` uses some heuristics to enable the HTTP keep-alive optimization, which means a single thread can serve multiple requests. Without detaching the context from the thread after serving a request, the initial request context is re-used for subsequent requests, which means propagated OTel context will be inherited by those request spans.

I tested this small patch in our production application and it appears to have resolved our APM woes.

### DIFF
--- a/instrumentation/wai/src/OpenTelemetry/Instrumentation/Wai.hs
+++ b/instrumentation/wai/src/OpenTelemetry/Instrumentation/Wai.hs
@@ -21,7 +21,7 @@ import Data.IP (fromHostAddress, fromHostAddress6)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Vault.Lazy as Vault
-import GHC.Stack (HasCallStack, callStack, popCallStack)
+import GHC.Stack (HasCallStack)
 import Network.HTTP.Types
 import Network.Socket
 import Network.Wai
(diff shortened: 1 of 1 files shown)


velveteer:
I believe that stripping W3C Trace Context headers can mask this issue. Which makes me wonder if this is related to https://www.iankduncan.com/articles/2023-08-28-opentelemetry-gotchas-phantom-spans.

iand675:
Can you rebase this against `main`? I think this change makes sense and would like to merge it.

velveteer:
Not sure if you were notified but I rebased this.

michaelpj (approved):
This makes sense, but please record the explanation from the PR in a comment, so future people will know what's going on.

velveteer:
Thanks @michaelpj. I've updated the commit message and added a code comment co-located with the change.

michaelpj:
Thanks!

PR #117, opened 2024-05-15
Add Datadog as a Known Propagator

pbrisbin:


pbrisbin:
Apologies if this is not the right place to discuss, but I'm noticing something as we continue to use OTel with Datadog...

It seems with the usual otel-collector and datadog exporter, the Traces show up with base16 TraceIds, not the numeric `Word64` values:



But correlation with logs that include the numeric `trace_id` still works!



Notice the search is for `trace_id:<base16>` but the found log has `trace_id:<word64>`.

It'll take be a bit to test if correlation still works when both the traces and logs use base16 values, but I kind of expect it to. If that's the case, do we need this conversion to word64 at all, or should we just put the base16 values in the headers when propagating?

pbrisbin:
Thanks, I can confirm that using hex trace-ids everywhere works great with Datadog. Here's an example I happen to have handy:



Traces show up in the UI using hex, then (6) logs, which are also using hex values (in structured JSON), are found just fine.

pbrisbin:
After verifying that ^ I kind of don't need this PR myself.

iand675:
Not much harm adding it either. Thanks for the contribution!

Issue #103, opened 2023-12-12
Support changes in http instrumentation field names


lf-:
Upstream otel specs have planned renaming a bunch of http instrumentation fields, and there's a migration plan here: https://opentelemetry.io/blog/2023/http-conventions-declared-stable/#migration-plan

PR #118, opened 2024-05-22
Update HTTP Semantic Conventions to Stable Versions

evanlauer1:
# Context
The OTEL spec declared the new HTTP semantic conventions as stable, so this PR is updating several instrumentation libraries to follow the new convenions. The updated libraries are http-client, wai, postgresql-simple, persistent-mysql, and yesod. Programmers can opt into the new conventions using the environment variable `OTEL_SEMCONV_STABILITY_OPT_IN`. The stable conventions will replace the old conventions in the next major release of hs-opentelemetry.

**Related:** Issue #103 

## Testing
- Created and ran test suite for semConvStabilityOptIn.hs to check if the program correctly reads the environment variable based on [this specification.](https://opentelemetry.io/docs/specs/semconv/http/)
- Did not test the changed instrumentation libraries to see if they add the correct attributes. The instrumentation libraries do not have any tests at present.


instrumentation/wai/hs-opentelemetry-instrumentation-wai.cabal
evanlauer1, line 1: These changes are auto-generated

api/hs-opentelemetry-api.cabal
evanlauer1, line 1: Auto-generated changes. Maybe `.cabal` and other auto-generated files should be in `.gitignore`?

hie.yaml
evanlauer1, line 1: This file was generated using the command `gen-hie > hie.yaml`. These changes allow HLS to work well with VsCode in this repository.

instrumentation/http-client/hs-opentelemetry-instrumentation-http-client.cabal
evanlauer1, line 1: Auto-generated change

instrumentation/http-client/test/Spec.hs
evanlauer1, line 1: Formatting

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
```diff
@@ -114,17 +121,32 @@ openMySQLConn tp attrs ci@MySQL.ConnectInfo {connectUser, connectPort, connectOp
                   MySQL.Pipe -> "pipe"
                   MySQL.Memory -> "inproc"
               _ -> Last Nothing
-    -- "net.sock.family" is unnecessary because it must be "inet" when "net.sock.peer.addr" or "net.sock.host.addr" is set.
-    attrs' =
+    addStableAttributes =
       H.union
-        [ ("db.connection_string", fromString $ showsPrecConnectInfoMasked 0 ci "")
+        [ ("db.connection_string" :: Text, fromString $ showsPrecConnectInfoMasked 0 ci "")
```
evanlauer1, line 126: `:: Text` is used because of `OverloadedStrings`

instrumentation/wai/hs-opentelemetry-instrumentation-wai.cabal
evanlauer1, line 1: Auto-generated

instrumentation/wai/package.yaml
```diff
@@ -28,6 +28,7 @@ dependencies:
 - bytestring
 - network
 - iproute
+- unordered-containers
```
evanlauer1, line 31: for Data.HashMap.Strict

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient/Raw.hs
```diff
@@ -107,27 +138,52 @@ instrumentResponse
...
-      s
-      [ ("http.status_code", toAttribute $ statusCode $ responseStatus resp)
-      -- TODO
-      -- , ("http.request_content_length",	_)
-      -- , ("http.request_content_length_uncompressed",	_)
-      -- , ("http.response_content_length", _)
-      -- , ("http.response_content_length_uncompressed", _)
-      -- , ("net.transport")
-      -- , ("net.peer.name")
-      -- , ("net.peer.ip")
```
evanlauer1, line 126: `net.peer.ip` is deleted because it is superseded by `server.address` which can either contain the contents of `net.peer.name` or `net.peer.ip`.

instrumentation/wai/src/OpenTelemetry/Instrumentation/Wai.hs
```diff
@@ -52,48 +61,114 @@ newOpenTelemetryWaiMiddleware' tp =
...
+      semConvStabilityOptIn <- getSemConvStabilityOptIn
+      let args =
+            defaultSpanArguments
+              { kind = Server
+              , attributes =
+                  case semConvStabilityOptIn of
+                    Stable ->
+                      usefulCallsite
+                        `H.union` [
+                                    ( "user_agent.original"
```
evanlauer1, line 73: This attribute must be provided to the sampler and is therefore has to be added when the span is created and not after.

evanlauer1 (commented):
Finished self-review.

instrumentation/persistent-mysql/src/OpenTelemetry/Instrumentation/Persistent/MySQL.hs
andrewsmith, line 1: I was surprised to see changes here, particularly special-casing on the same OTEL_SEMCONV_STABILITY_OPT_IN values. On the face of it, those seem unrelated. However, upon looking at the [database conventions](https://opentelemetry.io/docs/specs/semconv/database/), they reference this. It appears that the overall changes aren't HTTP-only. So it seems like you are doing the correct thing according to the spec.

andrewsmith (commented):
Nice work so far, Evan! This looks like excellent progress and is very thoughtful. I'll defer to Ian for approval, but this appears to be going in a good direction to me.

api/src/OpenTelemetry/SemConvStabilityOptIn.hs
```diff
@@ -0,0 +1,27 @@
...
+  let menvs = fmap T.strip . T.splitOn "," . T.pack <$> menv
+  pure $ case menvs of
+    Nothing -> Old
+    Just envs ->
+      if "http/dup" `elem` envs
+        then Both
+        else
+          if "http" `elem` envs
+            then Stable
+            else Old
```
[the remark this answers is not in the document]
MaxGabriel: I recommend multiway if for this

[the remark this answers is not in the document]
matto-mercury: you could use guards, I suppose:
```
Just envs | "http/dup" `elem` envs = Both
          | "http" `elem` envs = Stable
          | otherwise = Old
```

api/test/OpenTelemetry/SemConvStabilityOptInSpec.hs
```diff
@@ -0,0 +1,42 @@
...
+      semConvStabilityOptIn <- getSemConvStabilityOptIn
+      semConvStabilityOptIn `shouldBe` Both
+    it "is 'Both' when env var is 'http,http/dup'" $ do
+      setEnv envVarName "http,http/dup"
+      semConvStabilityOptIn <- getSemConvStabilityOptIn
+      semConvStabilityOptIn `shouldBe` Both
+    it "is 'Both' when env var is 'http,something-random,http/dup'" $ do
+      setEnv envVarName "http,http/dup"
+      semConvStabilityOptIn <- getSemConvStabilityOptIn
+      semConvStabilityOptIn `shouldBe` Both
```
MaxGabriel, line 42: I think your way is preferable, just throwing out an _option_ is to do these tests as a loop through `[("testname", "envvarname", valueItShouldBe), ...]`

instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
```diff
@@ -210,20 +217,31 @@ openTelemetryYesodMiddleware
...
+                  r <- mr
+                  pure ("http.route", toAttribute $ pathRender rr r)
+              , do
+                  r <- mr
+                  pure ("http.handler", toAttribute $ nameRender rr r)
+              , do
+                  ff <- lookup "X-Forwarded-For" $ requestHeaders req
+                  case semConvStabilityOptIn of
+                    Stable -> pure ("client.address", toAttribute $ T.decodeUtf8 ff)
+                    Both -> pure ("client.address", toAttribute $ T.decodeUtf8 ff)
```
MaxGabriel, line 236: imo it's clearer to use `Just` here. More concrete so it's more obvious what it's doing (vs what monad am I in for `pure`?) and contrasts better with Nothing on the next line

V minor though don't think you need to change this

instrumentation/http-client/test/Spec.hs
evanlauer1: Fixed

api/src/OpenTelemetry/SemConvStabilityOptIn.hs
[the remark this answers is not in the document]
evanlauer1: Done

api/src/OpenTelemetry/SemConvStabilityOptIn.hs
```diff
@@ -0,0 +1,27 @@
...
+) where
+
+import qualified Data.Text as T
+import System.Environment (lookupEnv)
+
+
+data SemConvStabilityOptIn = Stable | Both | Old deriving (Show, Eq)
+
+
+getSemConvStabilityOptIn :: IO SemConvStabilityOptIn
```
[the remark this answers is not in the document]
evanlauer1: I split it into two functions, but I am thinking about keeping the tests the same. I like that the current tests make sure that the program is reading the env var correctly, and the fact that `getSemConvStabilityOptIn` is trivial means that it is still a good test for `parseSemConvStabilityOptIn`. It might be better to test the pure function though, not sure.

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient/Raw.hs
```diff
@@ -64,36 +64,67 @@ instrumentRequest
...
+              ( "user_agent.original"
+              , toAttribute $ maybe "" T.decodeUtf8 (lookup hUserAgent $ requestHeaders req)
+              )
+            ]
+          addAttributes s
+            $ H.fromList
+            $ mapMaybe
+              (\h -> (\v -> ("http.request.header." <> T.decodeUtf8 (foldedCase h), toAttribute (T.decodeUtf8 v))) <$> lookup h (requestHeaders req))
+            $ requestHeadersToRecord conf
+        addOldAttributes = do
```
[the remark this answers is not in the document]
evanlauer1: Fixed

instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
evanlauer1: Changed it.

api/src/OpenTelemetry/SemConvStabilityOptIn.hs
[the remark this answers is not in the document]
matto-mercury: your tests are pretty nice, compact and legible, but one thing you could do with the pure function is reduce the number of env var tests to two (set/unset), and replace the rest with tests of the pure function that could be one-liners like `parseSemConvStabiltyOptIn (Just "http") `shouldBe` Stable`. if the setup was more onerous than `setEnv "blah"` I might feel more strongly about suggesting that approach, but as it stands I think it's just an aesthetic choice.

api/test/OpenTelemetry/SemConvStabilityOptInSpec.hs
evanlauer1: Done. I was on the edge but I think it increases readability to do it that way because there is less repeated information to look through.

api/src/OpenTelemetry/SemConvStabilityOptIn.hs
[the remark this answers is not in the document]
evanlauer1: I updated the tests to use a looping structure like Max suggested. Does that satisfy you're aesthetic preference?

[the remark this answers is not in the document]
matto-mercury: that's good too :)

iand675 (changes requested):
The direction here is really great! Left some feedback on how I'd evolve the interface a bit to future-proof it.

Issue #102, opened 2023-11-03
Combining WAI and Yesod instrumentation loses exception details


spencerjanssen:
I've got an application instrumented with both hs-opentelemetry-instrumentation-wai and hs-opentelemetry-instrumentation-yesod.  For the most part this works well, however I've found that the details of exceptions (exception message, exception backtrace, span status message) thrown by handlers never make it to WAI middleware trace.

I see why this happens: by default Yesod catches the exception and converts it to an error page -- the WAI instrumentation never sees the exception.  However, I'm not sure what the proper fix is.  The Yesod middleware could simply `recordException` targeting the WAI trace.  Unfortunately this still results in the WAI middleware setting a `""` span status message rather than a  meaningful description.

pbrisbin:
Hi there, I'm bumping into this currently.

>  The Yesod middleware could simply `recordException` targeting the WAI trace.

Based on my limited understanding, I think this may be the only option here.

>  Unfortunately this still results in the WAI middleware setting a `""` span status message rather than a meaningful description.

Can it not also use `setStatus` to address this?

I'm picturing something like...

```hs
instance Yesod App where
  -- ...

  yesodMiddleware h = h `onException` \ex -> do
    req <- waiRequest
    for_ (lookupSpan =<< requestContext req) $ \span -> do
      recordException span mempty Nothing ex
      setStatus span $ Error $ pack $ displayException ex
```

PR #121, opened 2024-05-30
Set exception details when using WAI and Yesod instrumentation together

pbrisbin:
### [Address warnings and HLint](https://github.com/iand675/hs-opentelemetry/pull/121/commits/23743a3297e67078e12684ca7115011f6cc4296d)
[23743a3](https://github.com/iand675/hs-opentelemetry/pull/121/commits/23743a3297e67078e12684ca7115011f6cc4296d)

### [Extract isInternalError](https://github.com/iand675/hs-opentelemetry/pull/121/commits/0d9cc2e336c39a4bd633848dcdb8aaef652cf0ce)
[0d9cc2e](https://github.com/iand675/hs-opentelemetry/pull/121/commits/0d9cc2e336c39a4bd633848dcdb8aaef652cf0ce)

### [Set exception details when using WAI and Yesod instrumentation together](https://github.com/iand675/hs-opentelemetry/pull/121/commits/9610e0fea06c548f3bc159718f30ce73a6f83df3)
[9610e0f](https://github.com/iand675/hs-opentelemetry/pull/121/commits/9610e0fea06c548f3bc159718f30ce73a6f83df3)

When the WAI instrumentation is used (so there is a span on the
request), this instrumentation only sets the shared attributes. Since
exceptions never reach WAI when used with Yesod (since they're caught to
error pages), this means that error traces don't include any details. By
calling `recordException` and `setStatus` here, we can address that.

Fixes https://github.com/iand675/hs-opentelemetry/issues/102

### DIFF
--- a/instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
+++ b/instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
@@ -1,6 +1,6 @@
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
@@ -261,8 +261,10 @@ openTelemetryYesodMiddleware rr m = do
           recordException waiSpan mempty Nothing ex
           setStatus waiSpan $ Error $ T.pack $ displayException ex
 
+
 shouldMarkException :: SomeException -> Bool
 shouldMarkException = maybe True isInternalError . fromException
+
 
 -- We want to mark the span as an error if it's an InternalError, the other
 -- HCError values are 4xx status codes which don't really count as a server
--- a/instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
+++ b/instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
@@ -256,6 +256,10 @@ openTelemetryYesodMiddleware rr m = do
         Right normal -> pure normal
     Just waiSpan -> do
       addAttributes waiSpan sharedAttributes
+
+      -- Explicitly record any exceptions here. When Yesod is in use, exceptions
+      -- are handled as error-pages before reaching the WAI middleware's inSpan,
+      -- meaning the exception details would not be otherwise attached.
       withException m $ \ex -> do
         when (shouldMarkException ex) $ do
           recordException waiSpan mempty Nothing ex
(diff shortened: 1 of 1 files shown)


pbrisbin:
Regarding the pre-commit hook... When I run foumolu I get a configuration error:

```console
% stack install --copy-compiler-tool fourmolu
Copying from
/home/patrick/.stack/snapshots/x86_64-linux-tinfo6/a76d5db647918a4cbcac406bf469be85d6333b9d6435f4dc08f2e0708720d750/9.2.7/bin/fourmolu
to /home/patrick/.stack/compiler-tools/x86_64-linux-tinfo6/ghc-9.2.7/bin/fourmolu.
Copied executables to /home/patrick/.stack/compiler-tools/x86_64-linux-tinfo6/ghc-9.2.7/bin/:
* fourmolu
% stack exec -- fourmolu --version
fourmolu 0.9.0.0 UNKNOWN UNKNOWN
using ghc-lib-parser 9.2.7.20230228
% stack exec -- fourmolu -i .
Failed to load /home/patrick/code/freckle/hs-opentelemetry/fourmolu.yaml:
Aeson exception:
Error in $['in-style']: unknown value: "no-space"
Valid values are: "left-align" or "right-align"


% 255 
```

instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
```diff
@@ -249,18 +247,29 @@ openTelemetryYesodMiddleware rr m = do
...
-            _ -> pure ()
+          when (isInternalError e) $ throwIO e
           pure (Left (e :: HandlerContents))
       case eResult of
         Left hc -> throwIO hc
         Right normal -> pure normal
     Just waiSpan -> do
       addAttributes waiSpan sharedAttributes
-      m
+      withException m $ \ex -> do
```
michaelpj, line 259: This makes sense, but please add the explanation from the PR here as a comment so we know why we're doing this in future.

michaelpj (approved):
If you update this I'll merge it.

instrumentation/yesod/src/OpenTelemetry/Instrumentation/Yesod.hs
pbrisbin: > so we know why we're doing this in future.

The PR description was taking from the [commit messages](https://github.com/iand675/hs-opentelemetry/pull/121/commits/9610e0fea06c548f3bc159718f30ce73a6f83df3), which means the information will be readily available and immutably attached to the code, which is why I typically prefer that over comments.

I will add the comment too, if you prefer, I just wanted to double check you knew it was in the commit message already too.

michaelpj: Commit messages become harder to find over time as the code changes in other ways. I would prefer a comment. (But I am an extremist: https://www.michaelpj.com/blog/2022/04/24/on-commenting-code.html)

pbrisbin: Done.

michaelpj:
Thanks!

Issue #119, opened 2024-05-23
Support OTLP 1.x


mx-psi:
Based on the submodule named 'proto' on the top-level folder of this repository, it looks like the OTLP proto definitions are based on open-telemetry/opentelemetry-proto@b43e9b18b76abf3ee040164b55b9c355217151f3, i.e. v0.11.0 from October 2021. This means that the current generated datatypes (those in the 'otlp' folder) reference fields that no longer exist in more recent OTLP versions.

For example, I can find this reference to `IntGauge`:

https://github.com/iand675/hs-opentelemetry/blob/02653bbfa286579be13cef17701da454ac3dfce1/otlp/src/Proto/Opentelemetry/Proto/Metrics/V1/Metrics.hs#L55

which was removed in v0.12.0 (specifically on open-telemetry/opentelemetry-proto/pull/342).

OTLP has since released a stable version, and some OpenTelemetry components no longer support v0.11.0 (e.g. [the OpenTelemetry Collector dropped support for OTLP v0.11.0 in v0.44.0](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CHANGELOG.md#v0440-beta)). In order to make this library work with the rest of the OpenTelemetry ecosystem, it would be necessary to bump the proto definitions

iand675:
Thanks for raising this!

michaelpj:
Fixed in https://github.com/iand675/hs-opentelemetry/pull/150

PR #122, opened 2024-05-30
Change OTLP generation strategy from manual to auto-generation

evanlauer1:
# Context
To resolve issue #119, it is preferable to auto-generate `.hs` files from the `.proto` files in the `proto` submodule instead of manually generating them as has been done in the past. This approach uses the strategy outlined [here](https://github.com/google/proto-lens?tab=readme-ov-file#readme), but using stack instead of cabal. 

## Auto-generation Instructions
To generate `.hs` files from a new version of the `proto` submodule, check that all of the modules that should be generated are listed under `generated-other-modules` in `otlp/package.yaml` and run `stack build`.

## Testing
- `stack build` compiles `hs-opentelemetry` without revisions to any packages except `hs-opentelemetry-otlp`
- `stack sdist otlp` when run from the root directory produces expected results
- The auto-generated files are present in the `otlp/.stack-work/build/autogen/Proto` directory after running `stack build`


otlp/hs-opentelemetry-otlp.cabal
evanlauer1, line 1: Auto-generated from `otlp/package.yaml`

otlp/package.yaml
```diff
@@ -8,6 +8,7 @@ version:             0.0.1.0
 extra-source-files:
 - README.md
 - ChangeLog.md
+- proto/**/*/*.proto
```
evanlauer1, line 11: Matches all `.proto` files in the `proto` submodule. The `proto` submodule was moved from the root directory to `otlp` because `hs-opentelemetry-otlp` needs it to auto-generate `.hs` files and no other packages use it.

otlp/proto
evanlauer1, line 1: I am unsure whether the original `proto` folder in the root directory should be deleted because of the Bugs section of [this article](https://git-scm.com/docs/git-mv). At present, the pull request does not delete this folder.

evanlauer1 (commented):
Finished self-review

PR #125, opened 2024-06-04
Rename `Log` to `LogRecord` and update data type to match logs data model spec

evanlauer1:
# Context
[The `Log` data model](https://opentelemetry.io/docs/specs/otel/logs/data-model/) has been updated since the `Log` data type was written. This PR renames [`Log` to `LogRecord`](https://opentelemetry.io/docs/concepts/glossary/#log) and updates `LogRecord` to match the current iteration of the `Log` data model. The largest change is that the current spec changed the type of the record field, `attributes`, from [Standard Attribute](https://opentelemetry.io/docs/specs/otel/common/#standard-attribute) to `map<string, any>`.

## Testing
- `stack build` runs

### DIFF
--- a/api/src/OpenTelemetry/LogAttributes.hs
+++ b/api/src/OpenTelemetry/LogAttributes.hs
@@ -1,0 +1,173 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE InstanceSigs #-}
+
+module OpenTelemetry.LogAttributes (
+  LogAttributes,
+  emptyAttributes,
+  addAttribute,
+  addAttributes,
+  getAttributes,
+  lookupAttribute,
+  AnyValue (..),
+  ToValue (..),
+
+  -- * unsafe utilities
+  unsafeLogAttributesFromListIgnoringLimits,
+  unsafeMergeLogAttributesIgnoringLimits,
+) where
+
+import Data.ByteString (ByteString)
+import Data.Data (Data)
+import qualified Data.HashMap.Strict as H
+import Data.Hashable (Hashable)
+import Data.Int (Int64)
+import Data.String (IsString (..))
+import Data.Text (Text)
+import qualified Data.Text as T
+import GHC.Generics (Generic)
+import OpenTelemetry.Attributes (AttributeLimits (..))
+
+
+data LogAttributes = LogAttributes
+  { attributes :: !(H.HashMap Text AnyValue)
+  , attributesCount :: {-# UNPACK #-} !Int
+  , attributesDropped :: {-# UNPACK #-} !Int
+  }
+  deriving stock (Show, Eq)
+
+
+emptyAttributes :: LogAttributes
+emptyAttributes = LogAttributes mempty 0 0
+
+
+addAttribute :: (ToValue a) => AttributeLimits -> LogAttributes -> Text -> a -> LogAttributes
+addAttribute AttributeLimits {..} LogAttributes {..} !k !v = case attributeCountLimit of
+  Nothing -> LogAttributes newAttrs newCount attributesDropped
+  Just limit_ ->
+    if newCount > limit_
+      then LogAttributes attributes attributesCount (attributesDropped + 1)
+      else LogAttributes newAttrs newCount attributesDropped
+  where
+    newAttrs = H.insert k (maybe id limitLengths attributeCountLimit $ toValue v) attributes
+    newCount = H.size newAttrs
+{-# INLINE addAttribute #-}
+
+
+addAttributes :: (ToValue a) => AttributeLimits -> LogAttributes -> H.HashMap Text a -> LogAttributes
+addAttributes AttributeLimits {..} LogAttributes {..} attrs = case attributeCountLimit of
+  Nothing -> LogAttributes newAttrs newCount attributesDropped
+  Just limit_ ->
+    if newCount > limit_
+      then LogAttributes attributes attributesCount (attributesDropped + H.size attrs)
+      else LogAttributes newAttrs newCount attributesDropped
+  where
+    newAttrs = H.union attributes $ H.map toValue attrs
+    newCount = H.size newAttrs
+{-# INLINE addAttributes #-}
+
+
+getAttributes :: LogAttributes -> (Int, H.HashMap Text AnyValue)
+getAttributes LogAttributes {..} = (attributesCount, attributes)
+
+
+lookupAttribute :: LogAttributes -> Text -> Maybe AnyValue
+lookupAttribute LogAttributes {..} k = H.lookup k attributes
+
+
+limitLengths :: Int -> AnyValue -> AnyValue
+limitLengths limit (TextValue t) = TextValue (T.take limit t)
+limitLengths limit (ArrayValue arr) = ArrayValue $ fmap (limitLengths limit) arr
+limitLengths limit (HashMapValue h) = HashMapValue $ fmap (limitLengths limit) h
+limitLengths _ val = val
+
+
+{- | An attribute represents user-provided metadata about a span, link, or event.
+
+ 'Any' values are used in place of 'Standard Attributes' in logs because third-party
+ logs may not conform to the 'Standard Attribute' format.
+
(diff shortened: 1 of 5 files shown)


api/src/OpenTelemetry/LogAttributes.hs
evanlauer1, line 1: Most functions in this file are duplicates of those in `Attributes.hs`, but updated to use the `AnyValue` and `LogAttributes` types. There might be some way to generalize the types of the functions in`Attributes.hs` to accommodate `AnyValue` and `LogAttributes`, but copying them over and modifying them is simpler.

api/src/OpenTelemetry/LogAttributes.hs
```diff
@@ -0,0 +1,173 @@
...
+import Data.ByteString (ByteString)
+import Data.Data (Data)
+import qualified Data.HashMap.Strict as H
+import Data.Hashable (Hashable)
+import Data.Int (Int64)
+import Data.String (IsString (..))
+import Data.Text (Text)
+import qualified Data.Text as T
+import GHC.Generics (Generic)
+import OpenTelemetry.Attributes (AttributeLimits (..))
```
evanlauer1, line 33: There is no necessary difference in `AttributeLimits` between `Attributes.hs` and `LogAttributes.hs`, so they are imported instead of rewritten.

api/src/OpenTelemetry/LogAttributes.hs
```diff
@@ -0,0 +1,173 @@
...
+
+ data Foo = Foo
+
+ instance ToValue Foo where
+   toValue Foo = TextValue "Foo"
+
+ @
+-}
+class ToValue a where
+  toValue :: a -> AnyValue
```
evanlauer1, line 127: Mirrors `ToAttribute` and `ToPrimitiveAttribute` in `Attributes.hs`.

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -2,29 +2,43 @@
...
+import OpenTelemetry.LogAttributes (LogAttributes)
 import OpenTelemetry.Resource (MaterializedResources)
 import OpenTelemetry.Trace.Id (SpanId, TraceId)
 
 
-data Log body = Log
+{- | This is a data type that can represent logs from various sources: application log files, machine generated events, system logs, etc. [Specification outlined here.](https://opentelemetry.io/docs/specs/otel/logs/data-model/)
+Existing log formats can be unambiguously mapped to this data type. Reverse mapping from this data type is also possible to the extent that the target log format has equivalent capabilities.
+-}
+data LogRecord body = LogRecord
```
evanlauer1, line 20: Renamed because of [this](https://opentelemetry.io/docs/concepts/glossary/#log).

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -2,29 +2,43 @@
...
   -- ^ Tuple contains three fields:
   --
   -- - Request trace id as defined in W3C Trace Context. Can be set for logs that are part of request processing and have an assigned trace id.
   -- - Span id. Can be set for logs that are part of a particular processing span.
   -- - Trace flag as defined in W3C Trace Context specification. At the time of writing the specification defines one flag - the SAMPLED flag.
   , severityText :: Maybe Text
-  -- ^ severity text (also known as log level). This is the original string representation of the severity as it is known at the source. If this field is missing and SeverityNumber is present then the short name that corresponds to the SeverityNumber may be used as a substitution. This field is optional.
+  -- ^ severity text (also known as log level). This is the original string representation of the severity as it is known at the source. If this field is missing
+  -- and SeverityNumber is present then the short name that corresponds to the SeverityNumber may be used as a substitution. This field is optional.
   , severityNumber :: Maybe Int64
```
evanlauer1, line 39: There are several possible [ways to represent `severityNumber`](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber) but `Int64` is the simplest for the moment

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -41,32 +55,40 @@ data Log body = Log
   -- +-----------------------+-------------+------------------------------------------------------------------------------------------+
   -- | 21-24                 | FATAL       | A fatal error such as application or system crash.                                       |
   -- +-----------------------+-------------+------------------------------------------------------------------------------------------+
-  , name :: Maybe Text
-  -- ^ Short low cardinality event type that does not contain varying parts. Name describes what happened (e.g. "ProcessStarted"). Recommended to be no longer than 50 characters. Typically used for filtering and grouping purposes in backends.
```
evanlauer1, line 45: The `name` field is no longer present in the spec

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -41,32 +55,40 @@ data Log body = Log
...
+  -- Smaller numerical values in each range represent less important (less severe) events. Larger numerical values in each range represent more important (more severe) events.
+  -- For example SeverityNumber=17 describes an error that is less critical than an error with SeverityNumber=20.
+  --
+  -- Mappings from existing logging systems and formats (or source format for short) must define how severity (or log level) of that particular format corresponds to SeverityNumber
+  -- of this data model based on the meaning given for each range in the above table. [More Information](https://opentelemetry.io/docs/specs/otel/logs/data-model/#mapping-of-severitynumber)
+  --
+  -- [These short names](https://opentelemetry.io/docs/specs/otel/logs/data-model/#displaying-severity) can be used to represent SeverityNumber in the UI
+  --
+  -- In the contexts where severity participates in less-than / greater-than comparisons SeverityNumber field should be used.
+  -- SeverityNumber can be compared to another SeverityNumber or to numbers in the 1..24 range (or to the corresponding short names).
```
evanlauer1, line 67: The above comments pertain to `severityNumber`

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -41,32 +55,40 @@ data Log body = Log
...
+  -- ^ A value containing the body of the log record. Can be for example a human-readable string message (including multi-line) describing the event in a free form or it can be a
+  -- structured data composed of arrays and maps of other values. Body MUST support any type to preserve the semantics of structured logs emitted by the applications.
+  -- Can vary for each occurrence of the event coming from the same source. This field is optional.
+  --
+  -- Type any (functions that use Log should have a Typeclass constraint of (ToValue body) => ...)
+  --    Value of type any can be one of the following:
+  --    - A scalar value: number, string or boolean,
+  --    - A byte array,
+  --    - An array (a list) of any values,
+  --    - A map<string, any>.
```
evanlauer1, line 78: Updated the formatting of the above comment

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -41,32 +55,40 @@ data Log body = Log
...
+  , resource :: Maybe MaterializedResources
+  -- ^ Describes the source of the log, aka resource. Multiple occurrences of events coming from the same event source can happen across time and they all have the same value of Resource.
+  -- Can contain for example information about the application that emits the record or about the infrastructure where the application runs. Data formats that represent this data model
+  -- may be designed in a manner that allows the Resource field to be recorded only once per batch of log records that come from the same source. SHOULD follow OpenTelemetry semantic conventions for Resources.
+  -- This field is optional.
+  , attributes :: LogAttributes
```
evanlauer1, line 84: Changed to the new `LogAttributes` type

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -92,7 +114,14 @@ data SeverityNumber
...
   | Unknown !Int32
-  deriving (Eq, Ord, Read, Show)
+  deriving (Eq, Ord, Read, Show )
+
+instance Enum SeverityNumber where
+  toEnum 1 = Trace
+  toEnum 2 = Trace2
+  toEnum 3 = Trace3
+  toEnum 4 = Trace4
+  ... -}
```
evanlauer1, line 124: One possible way to represent `severityNumber` is to create an Enum containing the short names (e.g. Fatal2, Trace3) and use `fromEnum` to retrieve an `Int` when needed. The benefit of this approach is that it is simple to pattern match on the short names and to use a `Show` instance to convert them to `String`s. I am still undecided if this is the best way to represent `severityNumber`, so it is still an `Int64` for now. Another option would be to use pattern synonyms defined on a `newtype` of `Int64`.

evanlauer1 (commented):
Finished self-review

iand675 (approved):

PR #126, opened 2024-06-08
Created emitLogRecord that takes a Logger as an argument. Refactored types to a separate module. stab-267

evanlauer1:
# Context
To fulfill the [OTel Logging Spec](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#emit-a-logrecord), `Logger`s must be able to emit `LogRecords`. This PR adds that basic capability and moves logging types to a separate module to avoid cyclic module dependencies.

## Testing
- `stack build` runs

PR #127, opened 2024-06-10
Added instrumentationScope to Logger and LogRecord

evanlauer1:
# Context
To comply with the spec, [`LogRecord` must have a field `instrumentationScope`](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-instrumentationscope). `Logger` provides this field to `LogRecord`s. This PR updates the codebase to match that spec. The `InstrumentationLibrary` data type was also [updated to include `schema_url` and `attributes`](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#get-a-logger). The haddock documentation about how to initialize an `instrumentationLibrary` has not yet been updated because I am unsure about how adding fields to `instrumentationScope` will affect things and whether it is necessary to add the fields.

## Testing
- `stack build` runs
- I am not sure how adding fields to `instrumentationScope` will affect things but it should maybe be tested

### DIFF
--- a/api/src/OpenTelemetry/Attributes.hs
+++ b/api/src/OpenTelemetry/Attributes.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE StrictData #-}
 
 {- |
@@ -75,7 +74,10 @@ data Attributes = Attributes
   , attributesCount :: {-# UNPACK #-} !Int
   , attributesDropped :: {-# UNPACK #-} !Int
   }
-  deriving stock (Show, Eq)
+  deriving stock (Show, Generic, Eq, Ord)
+
+
+instance Hashable Attributes
 
 
 emptyAttributes :: Attributes
--- a/api/src/OpenTelemetry/Attributes.hs
+++ b/api/src/OpenTelemetry/Attributes.hs
@@ -74,10 +74,7 @@ data Attributes = Attributes
   , attributesCount :: {-# UNPACK #-} !Int
   , attributesDropped :: {-# UNPACK #-} !Int
   }
-  deriving stock (Show, Generic, Eq, Ord)
-
-
-instance Hashable Attributes
+  deriving stock (Show, Eq, Ord)
 
 
 emptyAttributes :: Attributes
--- a/api/src/OpenTelemetry/Attributes.hs
+++ b/api/src/OpenTelemetry/Attributes.hs
@@ -74,7 +74,10 @@ data Attributes = Attributes
   , attributesCount :: {-# UNPACK #-} !Int
   , attributesDropped :: {-# UNPACK #-} !Int
   }
-  deriving stock (Show, Eq, Ord)
+  deriving stock (Show, Generic, Eq, Ord)
+
+
+instance Hashable Attributes
 
 
 emptyAttributes :: Attributes
--- a/api/src/OpenTelemetry/Internal/Common/Types.hs
+++ b/api/src/OpenTelemetry/Internal/Common/Types.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE InstanceSigs #-}
 
 module OpenTelemetry.Internal.Common.Types (InstrumentationLibrary (..)) where
@@ -46,7 +47,10 @@ data InstrumentationLibrary = InstrumentationLibrary
   , librarySchemaUrl :: {-# UNPACK #-} !Text
   , libraryAttributes :: Attributes
   }
-  deriving (Ord, Eq, Show)
+  deriving (Ord, Eq, Generic, Show)
+
+
+instance Hashable InstrumentationLibrary
 
 
 instance IsString InstrumentationLibrary where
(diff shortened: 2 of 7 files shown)


evanlauer1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/127?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#127** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#126** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/126?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

api/src/OpenTelemetry/Attributes.hs
```diff
@@ -75,7 +74,10 @@ data Attributes = Attributes
   , attributesCount :: {-# UNPACK #-} !Int
   , attributesDropped :: {-# UNPACK #-} !Int
   }
-  deriving stock (Show, Eq)
+  deriving stock (Show, Generic, Eq, Ord)
+
+
+instance Hashable Attributes
```
evanlauer1, line 80: Needed for `instance Hashable InstrumentationLibrary`

api/src/OpenTelemetry/Internal/Common/Types.hs
evanlauer1, line 1: This is a new type module for types used by both `OpenTelemetry.Internal.Trace.Types` and `OpenTelemetry.Internal.Logging.Types`

api/src/OpenTelemetry/Internal/Common/Types.hs
```diff
@@ -0,0 +1,56 @@
...
+
+ @
+-}
+data InstrumentationLibrary = InstrumentationLibrary
+  { libraryName :: {-# UNPACK #-} !Text
+  -- ^ The name of the instrumentation library
+  , libraryVersion :: {-# UNPACK #-} !Text
+  -- ^ The version of the instrumented library
+  , librarySchemaUrl :: {-# UNPACK #-} !Text
+  , libraryAttributes :: Attributes
```
evanlauer1, line 47: `librarySchemaUrl` and `libraryAttributes` were added to this data type because of [this spec for `Logger`s](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#get-a-logger).

api/src/OpenTelemetry/Internal/Trace/Types.hs
```diff
@@ -38,51 +39,6 @@ data ExportResult
   | Failure (Maybe SomeException)
 
 
-{- | An identifier for the library that provides the instrumentation for a given Instrumented Library.
```
evanlauer1, line 41: Moved to `OpenTelemetry.Internal.Common.Types`

api/src/OpenTelemetry/Internal/Common/Types.hs
```diff
@@ -0,0 +1,56 @@
...
+   }
+
+ @
+-}
+data InstrumentationLibrary = InstrumentationLibrary
+  { libraryName :: {-# UNPACK #-} !Text
+  -- ^ The name of the instrumentation library
+  , libraryVersion :: {-# UNPACK #-} !Text
+  -- ^ The version of the instrumented library
+  , librarySchemaUrl :: {-# UNPACK #-} !Text
```
evanlauer1, line 46: I'm not sure if `Text` is the best type to represent a `schemaUrl`. I think another data type is defined in `OpenTelemetry.Resource`.

evanlauer1 (commented):
Finished self-review.

api/src/OpenTelemetry/Attributes.hs
iand675: I can't think of a reason at this point where `InstrumentationLibrary` or this would benefit from supplying a `Hashable` instance.

iand675 (changes requested):

api/src/OpenTelemetry/Attributes.hs
evanlauer1: It looks like one of the Processors uses the Hashable instance of InstrumentationLibrary
Screenshot 2024-06-10 at 1 45 56 PM

iand675: Ah I see. The in the previous version, it was a name + version pair, which makes sense as a thing that's hashable. By adding attributes as well, you could technically end up with two `InstrumentationLibrary` values with the same name and version, but different attributes, which is a little bit odd here 🤔 . I think it would make sense to have the `InstrumentationLibrary` keep the shape prior to this PR, but have a `HashMap InstrumentationLibrary ...attributes...` that we use to add on attributes upon export.

evanlauer1: [From the docs](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#get-a-logger):
Screenshot 2024-06-10 at 4 37 16 PM

evanlauer1: It still might be worth making it impossible to make two loggers with different attributes but the same identity, but the docs says that is a user error so I'm not sure if we need to worry about that.

api/src/OpenTelemetry/Internal/Common/Types.hs
iand675: Text seems fine to me for now.

PR #128, opened 2024-06-11
Add LoggerProvider to create Loggers and allow registering a global LoggerProvider

evanlauer1:
# Context
Adds `LoggerProvider`s according to the [OTel spec](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#loggerprovider). Allows registering and getting a global `LoggerProvider`. Some methods of `LoggerProvider` like [`shutdown`](https://opentelemetry.io/docs/specs/otel/logs/sdk/#shutdown) will be implemented when the SDK is created.

## Testing
- `stack build` runs

### DIFF
--- a/api/src/OpenTelemetry/Logging/Core.hs
+++ b/api/src/OpenTelemetry/Logging/Core.hs
@@ -2,7 +2,18 @@
 {-# LANGUAGE TypeApplications #-}
 
 module OpenTelemetry.Logging.Core (
+  -- LoggerProvider operations
+  LoggerProvider (..),
+  LoggerProviderOptions (..),
+  emptyLoggerProviderOptions,
+  createLoggerProvider,
+  globalLoggerProvider,
+  setGlobalLoggerProvider,
+  getGlobalLoggerProvider,
+  -- Logger operations
   Logger (..),
+  makeLogger,
+  -- LogRecord operations
   LogRecord (..),
   LogRecordArguments (..),
   mkSeverityNumber,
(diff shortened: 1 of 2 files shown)


evanlauer1:
* **#128** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/128?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -15,19 +28,73 @@ import Control.Applicative
...
+
+ You should generally use @getGlobalLoggerProvider@ for most applications.
+-}
+createLoggerProvider :: (MonadIO m) => LoggerProviderOptions -> m LoggerProvider
+createLoggerProvider LoggerProviderOptions {..} = pure LoggerProvider {loggerProviderResource = loggerProviderOptionsResource}
+
+
+globalLoggerProvider :: IORef LoggerProvider
+globalLoggerProvider = unsafePerformIO $ newIORef =<< createLoggerProvider emptyLoggerProviderOptions
+{-# NOINLINE globalLoggerProvider #-}
```
evanlauer1, line 72: This `IORef` mirrors `globalTracer` in `OpenTelemetry.Trace.Core`

api/src/OpenTelemetry/Logging/Core.hs
evanlauer1, line 1: Most of these functions and types mirror those in `OpenTelemetry.Trace.Core`

evanlauer1 (commented):
Self-review completed.

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -15,19 +28,73 @@ import Control.Applicative
...
 getCurrentTimestamp :: (MonadIO m) => m Timestamp
 getCurrentTimestamp = liftIO $ coerce @(IO TimeSpec) @(IO Timestamp) $ getTime Realtime
 
 
+data LoggerProviderOptions = LoggerProviderOptions {loggerProviderOptionsResource :: MaterializedResources}
+
+
+{- | Options for creating a @LoggerProvider@ with no resources and default limits.
+
+ In effect, logging is a no-op when using this configuration.
```
andrewsmith, line 53: Trying to orient myself to this: what is it about this that makes logging a no-op? Is it that this is part of the API and thus not tied to a concrete implementation?

evanlauer1: That is a great question. I'm a little foggy about this idea myself, but I think that it means that it has no `Processors` or `Exporters` or any way to interact with the outside world. The API is supposed to be completely [no-op](https://opentelemetry.io/docs/specs/otel/logs/noop/) when the SDK has not registered a global LoggerProvider (See the second paragraph of `api/Readme.md`).

This got me thinking, so I'm working on making a way to make it clearer when things are no-op.

evanlauer1: Okay. Logging is no-op when the `LoggerProvider` has no processors. I think I will make a `noOpLoggerProvider` value in a later pull request to make this more clear.

iand675 (approved):

PR #129, opened 2024-06-11
Refactored attribute limits to be controlled by the LoggerProvider

evanlauer1:
# Context
[`LoggerProvider`s should control the `AttributeLimit`s for `Logger`s](https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits). Previously, the entirety of `LogAttributes` was provided to the `emitLogRecord` function instead of passing the attributes via a `HashMap Text AnyValue` and `AttributeLimit`s via a `LoggerProvider` accessed through the provided `Logger`. Relatedly, [`LogAttributes` should be able to be changed after the creation of the `Logger`](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord), but that will be added later.

## Testing
- `stack build` runs

### DIFF
--- a/api/src/OpenTelemetry/Logging/Core.hs
+++ b/api/src/OpenTelemetry/Logging/Core.hs
@@ -28,15 +28,18 @@ import Control.Applicative
 import Control.Monad.Trans
 import Control.Monad.Trans.Maybe
 import Data.Coerce
+import qualified Data.HashMap.Strict as H
 import Data.IORef
 import Data.Maybe
 import GHC.IO (unsafePerformIO)
+import OpenTelemetry.Attributes (AttributeLimits)
 import OpenTelemetry.Common
 import OpenTelemetry.Context
 import OpenTelemetry.Context.ThreadLocal
 import OpenTelemetry.Internal.Common.Types
 import OpenTelemetry.Internal.Logging.Types
 import OpenTelemetry.Internal.Trace.Types
+import OpenTelemetry.LogAttributes
 import OpenTelemetry.Resource (MaterializedResources, emptyMaterializedResources)
 import System.Clock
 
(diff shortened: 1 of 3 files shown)


evanlauer1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/129?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#129** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/129?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#128** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -124,5 +130,9 @@ emitLogRecord Logger {..} LogRecordArguments {..} = do
       , logRecordBody = body
       , logRecordResource = loggerProviderResource loggerProvider
       , logRecordInstrumentationScope = loggerInstrumentationScope
-      , logRecordAttributes = attributes
+      , logRecordAttributes =
+          addAttributes
+            (loggerProviderAttributeLimits loggerProvider)
+            emptyAttributes
+            attributes
```
andrewsmith, line 137: Is there anything about this that addresses the part of the spec that says this:
> There SHOULD be a message printed in the SDK’s log to indicate to the user that an attribute was discarded due to such a limit. To prevent excessive logging, the message MUST be printed at most once per LogRecord (i.e., not per discarded attribute).

iand675: This is actually something that we don't support on the tracing side (which says effectively the same thing). I wanted to support it on the tracing side using OTel logging, so it's a bit of an ouroboros 😆 

I think enough of the logging interface is in place that we should be be able to support this use case now though.

evanlauer1: This is addressed by the next upstream pull request in the stack. The change was to large to be reasonably contained in this PR.

iand675 (approved):

PR #131, opened 2024-06-12
Refactored LogRecords to be mutable so that attributes can be modified after creation

evanlauer1:
# Big Context
Logging Support is being added to hs-opentelemetry. [Logging spec](https://opentelemetry.io/docs/specs/otel/logs/)

## Small (This PR) Context
[`LogRecord`s should be modifiable](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord) after creation for the [purposes of `LogRecordProcessor`s](https://opentelemetry.io/docs/specs/otel/logs/sdk/#onemit). This PR refactors `LogRecord` to contain a mutable `IORef` to an `ImmutableLogRecord`. It also adds convenience functions for adding attributes to `LogRecord`s and a test suite that tests `LoggerProvider`s and `LogRecord`s. 

`LogRecord`s can only be read through the functions of `ReadableLogRecord` and modified by the functions of `ReadWriteLogRecord`. [This is in accordance with the spec](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readablelogrecord).

The type of the `body` field was changed from a type variable to an `AnyValue` so that processors and exporters can process logs from various sources that might have different body types.

## Testing
- `stack build` runs
- New test suite under `stack test hs-opentelemetry` passes

PR #133, opened 2024-07-02
Added Logging when creating spans or emitting logs drops attributes. Updated Severity Number

evanlauer1:
# Big Context
Logging Support is being added to hs-opentelemetry. [Logging spec](https://opentelemetry.io/docs/specs/otel/logs/)

## Small (This PR) Context
If creating a `Span` drops attributes, [a warning is now logged](https://opentelemetry.io/docs/specs/otel/trace/sdk/#span-limits). The same goes for [emitting logs](https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecord-limits). `SeverityNumber` was updated to be an `Enum` instead of a datatype containing both a number and a name. Helper functions were added for emitting logs from the `hs-opentelemetry` library.

## Testing
- `stack build` runs
-  Test suite passes


evanlauer1:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/133?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#133** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/133?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#129** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/129?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#131](https://github.com/iand675/hs-opentelemetry/pull/131) <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/131?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#128** <a href="https://app.graphite.dev/github/pr/iand675/hs-opentelemetry/128?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

api/src/OpenTelemetry/Internal/Common/Types.hs
```diff
@@ -30,7 +30,7 @@ import OpenTelemetry.Attributes (Attributes, emptyAttributes)
  import qualified Data.Text as T
  import Data.Version (showVersion)
  import OpenTelemetry.Attributes (emptyAttributes)
- import Paths_your_package_name
+ import Paths_your_package_name (version)
```
evanlauer1, line 33: Makes it easier to see where `version` comes from.

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
   , context :: Maybe Context
   , severityText :: Maybe Text
-  , severityNumber :: Maybe Int
+  , severityNumber :: Maybe SeverityNumber
   , body :: body
   , attributes :: H.HashMap Text AnyValue
   }
 
 
-data SeverityNumber = SeverityNumber {shortName :: Maybe Text, severityInt :: !Int64} deriving (Read, Show)
```
evanlauer1, line 135: This data structure is a bit confusing. It is odd that it contains both `severityInt` which is the actual data being represented and also `shortName` which is kind of Metadata about the number. An `Enum` seems less overcomplicated.

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
+  | Info3
+  | Info4
+  | Warn
+  | Warn2
+  | Warn3
+  | Warn4
+  | Err
+  | Err2
+  | Err3
+  | Err4
```
evanlauer1, line 168: `Err` is used because `Error` is taken by another data structure in the haskell ecosystem that was causing naming conflicts. There may be a better name for this but `Err` seems fine.

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
+  | Fatal
+  | Fatal2
+  | Fatal3
+  | Fatal4
+  | Unknown !Int
+
+
+instance Enum SeverityNumber where
```
evanlauer1, line 176: This is explicitly declared so there is no doubt about which `Int` each `SeverityNumber` correlates to. Also, `Unknown` needs to be handled as a special case.

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
+  fromEnum Fatal2 = 22
+  fromEnum Fatal3 = 23
+  fromEnum Fatal4 = 24
+  fromEnum (Unknown n) = n
 
 
 instance Eq SeverityNumber where
-  (==) :: SeverityNumber -> SeverityNumber -> Bool
-  n == m = severityInt n == severityInt m
+  (==) = on (==) fromEnum
```
evanlauer1, line 232: `on` makes this more concise at the possible cost of understandability. I think `on` describes the concept pretty well though.

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
-mkShortName 20 = Just "ERROR4"
-mkShortName 21 = Just "FATAL"
-mkShortName 22 = Just "FATAL2"
-mkShortName 23 = Just "FATAL3"
-mkShortName 24 = Just "FATAL4"
-mkShortName _ = Nothing
+  compare = on compare fromEnum
+
+
+toShortName :: SeverityNumber -> Maybe Text
```
evanlauer1, line 239: This is used instead of `Show` because the `String`s/`Text` don't directly correspond to the `Enum` because of capitalization and differences between `Err` and `"ERROR"`.

api/src/OpenTelemetry/Internal/Trace/Types.hs
```diff
@@ -87,7 +87,6 @@ data TracerProvider = TracerProvider
   , tracerProviderAttributeLimits :: !AttributeLimits
   , tracerProviderSpanLimits :: !SpanLimits
   , tracerProviderPropagators :: !(Propagator Context RequestHeaders ResponseHeaders)
-  , tracerProviderLogger :: LogRecord Text -> IO ()
```
evanlauer1, line 90: This is not the correct way to emit a `LogRecord`. This was a hypothetical implementation written before Logging was fleshed out.

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -120,19 +132,48 @@ emitLogRecord Logger {..} LogRecordArguments {..} = do
     SpanContext {traceId, spanId, traceFlags} <- getSpanContext currentSpan
     pure (traceId, spanId, traceFlags)
 
+  let logRecordAttributes =
+        LA.addAttributes
+          (loggerProviderAttributeLimits loggerProvider)
+          LA.emptyAttributes
+          attributes
+
+  when (LA.attributesDropped logRecordAttributes > 0) $ void logDroppedAttributes
```
evanlauer1, line 141: Logs a message that attributes were dropped.

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -120,19 +132,48 @@ emitLogRecord Logger {..} LogRecordArguments {..} = do
...
-      , logRecordAttributes =
-          addAttributes
-            (loggerProviderAttributeLimits loggerProvider)
-            emptyAttributes
-            attributes
+      , logRecordAttributes
+      }
+
+
+-- | WARNING: this function should only be used to emit logs from the hs-opentelemetry-api library. DO NOT USE this function in any other context.
```
evanlauer1, line 157: I'm not sure if this is the best way to communicate this. It would be ideal if these functions could be in an internal module that is not exported. That is not simple because it introduces a cyclic module dependency.

api/src/OpenTelemetry/Logging/Core.hs
```diff
@@ -120,19 +132,48 @@ emitLogRecord Logger {..} LogRecordArguments {..} = do
...
+logDroppedAttributes :: (MonadIO m) => m (LogRecord Text)
+logDroppedAttributes = emitOTelLogRecord H.empty Warn "At least 1 attribute was discarded due to the attribute limits set in the logger provider."
+
+
+-- | WARNING: this function should only be used to emit logs from the hs-opentelemetry-api library. DO NOT USE this function in any other context.
+emitOTelLogRecord :: (MonadIO m) => H.HashMap Text LA.AnyValue -> SeverityNumber -> Text -> m (LogRecord Text)
```
evanlauer1, line 163: Helper function to remove boilerplate. Takes `Attributes`, a `SeverityNumber`, and a message because those are the only things that need to be set in this context at this point. `Attributes` may be unnecessary, but it forces the person writing the code that emits an OTel log to think about whether there are `Attributes` that could be attached.

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -285,10 +286,13 @@ createSpanWithoutCallStack t ctxt n args@SpanArguments {..} = liftIO $ do
...
                     , spanTracer = t
                     }
+
+            when (A.attributesDropped (spanAttributes is) > 0) $ void logDroppedAttributes
+
             s <- newIORef is
             eResult <- try $ mapM_ (\processor -> processorOnStart processor s ctxt) $ tracerProviderProcessors $ tracerProvider t
             case eResult of
-              Left err -> print (err :: SomeException)
+              Left err -> void $ emitOTelLogRecord H.empty Err $ T.pack $ show (err :: SomeException)
```
evanlauer1, line 295: With the `print` statement, this would still be in effect "logged" even if logging is no-op. The code should probably be `grep`ed for print statements that print errors.

evanlauer1 (commented):
Finished self-review

api/src/OpenTelemetry/Internal/Logging/Types.hs
```diff
@@ -126,55 +126,139 @@ data LogRecordArguments body = LogRecordArguments
...
+    , observedTimestamp = Nothing
+    , context = Nothing
+    , severityText = Nothing
+    , severityNumber = Nothing
+    , body = body
+    , attributes = H.empty
+    }
+
+
+data SeverityNumber
```
evanlauer1, line 148: An `Enum` is used so that setting `severityNumber` in `LogRecordArguments` does not need an `Int` with little semantic information. It is much easier to decide what number to use now because they are well labeled.

hie.yaml
```diff
@@ -90,7 +90,7 @@ cradle:
     - path: "instrumentation/yesod/test"
       component: "hs-opentelemetry-instrumentation-yesod:test:hs-opentelemetry-instrumentation-yesod-test"
 
-    - path: "otlp/src"
+    - path: "otlp/./"
```
andrewsmith, line 93: Can this be normalized?

```suggestion
    - path: "otlp"
```

andrewsmith (commented):
Overall, this looks good to me! I just had a few small comments.

api/src/OpenTelemetry/Logging/Core.hs
evanlauer1: Tracing/Core.hs uses it

api/src/OpenTelemetry/Internal/Logging/Types.hs
evanlauer1: Did some poking around and it actually looks like [the conflict is coming from here](https://opentelemetry.io/docs/specs/otel/trace/api/#set-status). Unfortunate that they used `Error` in two parts of the Spec!

I feel like `SpanStatus` should be able to use `Error` because the spec explicitly uses that term whereas `SeverityNumber` defines the short name as `ERROR` but has no guidance on `Enum` names. It could be qualified, but I feel like that is more cumbersome than naming it `Err`.

hie.yaml
evanlauer1: This file was generated by `gen-hie`, so it could be changed but it's probably not necessary. It is likely that successive runs of `gen-hie` will continue to update it anyway.

iand675 (changes requested):
just realized that my comments on this were left unpublished, sorry 😭

hie.yaml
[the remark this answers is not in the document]
evanlauer1: I changed it to `otlp/`, re-ran `gen-hie > hie.yaml`, and it replaced it with `otlp/./` again. I will defer to `gen-hie` because every time we re-run `gen-hie` it will just update this again.

api/src/OpenTelemetry/Logging/Core.hs
evanlauer1: Fixed so that there is `OpenTelemetry.Internal.Logging.Core`

iand675 (approved):

PR #140, opened 2024-07-31
add protobuf nix dependency to stack.yaml

kakkun61:
This PR will add protobuf as Nix dependency because hs-opentelemetry-otlp requires `protoc` newly.

I need help to add the protobuf package to flake.nix. It is required by cabal-install users.


iand675 (approved):

kakkun61:
for cabal-install #143

PR #142, opened 2024-08-07
semantic conventions

kakkun61:
# What

- auto-generated attributes keys from https://github.com/open-telemetry/semantic-conventions/
- keys with their value types
- to name attribute maps which is without dropping (`Attributes` have dropping)

# Why

- OpenTelemetry has many predefined keys for attributes and manual implementation is too hard
- using `Attribute` causes a lot of pattern matchings even when programmers can identify its types uniquely
- both just maps and `Attributes` are often named as `attributes` variable names and it confuses

# How

- to make semantic-conventions package
  - its contents are auto-generated by _Setup.hs_
- a new data type `Key a` is introduced
  - `a` is a correspondent value's type
  - `insertByKey`, `lookupByKey` and so on
- to name `HashMap Text Attribute` as `AttributeMap`
  - relative functions

### DIFF
--- a/semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
+++ b/semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
@@ -1,0 +1,70 @@
+cabal-version: 2.4
+
+name: hs-opentelemetry-semantic-conventions
+version: 0.0.1.0
+author: Kazuki Okamoto (岡本和樹)
+maintainer: kazuki.okamoto@herp.co.jp
+synopsis: OpenTelemetry Semantic Conventions for Haskell
+description:
+  OpenTelemetry Semantic Conventions for Haskell is a library that is automatically generated
+  based on [semantic-conventions](https://github.com/open-telemetry/semantic-conventions/) v1.24.
+
+common common
+  build-depends: base >= 4 && < 5
+  ghc-options: -Wall
+  if impl(ghc >= 8.0)
+    ghc-options: -Wcompat
+  default-language: Haskell2010
+
+common artifact
+  import: common
+  ghc-options: -Wno-name-shadowing
+  if impl(ghc >= 6.4)
+    ghc-options: -Wincomplete-record-updates
+  if impl(ghc >= 6.8)
+    ghc-options: -Wmonomorphism-restriction
+  if impl(ghc >= 7.0)
+    ghc-options: -Wmissing-import-lists
+  if impl(ghc >= 7.2)
+    ghc-options: -Wincomplete-uni-patterns
+                 -Widentities
+  if impl(ghc >= 8.0)
+    ghc-options: -Wmissing-exported-signatures
+                 -Wredundant-constraints
+  if impl(ghc >= 8.2)
+    ghc-options: -Wmissing-home-modules
+  if impl(ghc >= 8.4)
+    ghc-options: -Wmissing-export-lists
+                 -Wpartial-fields
+  if impl(ghc >= 8.8)
+    ghc-options: -Wmissing-deriving-strategies
+  if impl(ghc >= 8.10)
+    ghc-options: -Wunused-packages
+  if impl(ghc >= 9.0)
+    ghc-options: -Winvalid-haddock
+  if impl(ghc >= 9.2)
+    ghc-options: -Wmissing-kind-signatures
+                 -Woperator-whitespace
+                 -Wredundant-bang-patterns
+
+library
+  import: artifact
+  hs-source-dirs: src
+  exposed-modules: OpenTelemetry.SemanticConventions
+  build-depends: hs-opentelemetry-api,
+                 text,
+
+executable generate
+  import: artifact
+  hs-source-dirs: dev
+  main-is: generate.hs
+  build-depends: aeson,
+                 base,
+                 Glob,
+                 directory,
+                 filepath,
+                 yaml,
+                 text,
+                 -- unordered-containers is necessary only when aeson < 2 is used
+                 unordered-containers,
+                 vector
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "semantic-conventions/model"]
+	path = semantic-conventions/model
+	url = https://github.com/open-telemetry/semantic-conventions/
--- a/cabal.project
+++ b/cabal.project
@@ -2,6 +2,7 @@ packages:
     api
   , sdk
   , otlp
+  , semantic-conventions
   , examples/hspec
   , examples/yesod-minimal
   , exporters/handle
--- a/flake.nix
+++ b/flake.nix
@@ -41,6 +41,7 @@
           ++ [
             "^otlp/src/"
             ".*\\.cabal$"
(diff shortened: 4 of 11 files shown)


kakkun61:
CI fails because of missing `git submodule update --init`.

How to fix?

kakkun61:
Git submodule should not be used?

👀 https://github.com/iand675/hs-opentelemetry/pull/122/commits/737f8229161e528fc63ec2b9b5d519e6b533769a

michaelpj:
As a non-maintainer:
- This looks interesting and useful.
- There seem to be a lot of unnecessary changes in this PR. I can't see why we need all the changes to the `Attributes` interface. At the end of the day, this is mainly producing a big list of `Text` keys, right? There's a lot here, I think it's a lot more likely to get reviewed if it doesn't have extra stuff.
- Personally, for this kind of rarely-changing code-generation, I think it is better to check in the result and have a script to re-generate it. 
    - Doing the generation in a custom setup means that the generation process needs to work flawlessly on any user's machine that builds the package. 
    - If it's just some generated files, we probably don't even need a separate folder, we could make do with a `generated` `hs-source-dir` for `hs-opentelemetry-api`.
    - It might be easier to review the generated files than the generator!
    - Plus custom setups are just generally a big pain.

michaelpj:
(I now have the right to approve and merge PRs, if that makes my comments seem more compelling :) )

kakkun61:
Some contents are moved to #157

michaelpj:
Do we want this to be a separate package rather than just a module in `hs-opentelemetry-api`?

kakkun61:
Users may want to use an individual version of hs-opentelemetry-semantic-conventions because they may want to use [an older version of semantic-convantions](https://github.com/open-telemetry/semantic-conventions/releases).

michaelpj:
That makes sense! 

I think I'd mildly prefer to use the same strategy for getting the git repository as https://github.com/iand675/hs-opentelemetry/pull/159 did: i.e. just pull it down as part of the generation script. Otherwise it's a bit inconsistent about what we do with submodules. WDYT @marinelli ?

semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
```diff
@@ -0,0 +1,73 @@
...
+name: hs-opentelemetry-semantic-conventions
+version: 1.24.0.0
+author: Kazuki Okamoto (岡本和樹)
+maintainer: kazuki.okamoto@herp.co.jp
+synopsis: OpenTelemetry Semantic Conventions for Haskell
+description:
+  OpenTelemetry Semantic Conventions for Haskell is a library that is automatically generated
+  based on [semantic-conventions](https://github.com/open-telemetry/semantic-conventions/) v1.24.
+
+  This module's version A.B.X.Y means that A.B is the version of the original semantic-conventions,
```
michaelpj, line 12:
```suggestion
  This package's version A.B.X.Y means that A.B is the version of the original semantic-conventions,
```

michaelpj: Note that this strategy doesn't let us do major releases of this package _without_ a corresponding version bump of the semantic conventions, right? That seems awkward. 

In the past I've mostly just given up on this approach and given the Haskell package its own versioning scheme. That's what we do with the OTLP package in this repository, for example.

kakkun61: Hmm how about $(A \times 100 + B).x.y.z$?

I think users want to find the newer version of hs-semconv that compatible with a semconv version. When hs-semconv has its own versioning, A conversion table is necessary to find compatible versions and also version bounds, like `= 1029.1.*`, cannot be used.

michaelpj: I guess I am unsure whether people actually really need what you suggest. My experience (e.g. with the LSP protocol package) has been that specification versions supported by a library change quite rarely, and people either tend to want "the latest" (in which case they don't care), or unusually "a very specific version", in which case they have to pin a version of the package anyway.

As for a conversion table, I think it would be good enough to just list the supported spec version in the package description? Then it doesn't take too long to click through the versions on Hackage to find the one that you want?

kakkun61: OK. Just follows PVP.

marinelli:
> That makes sense!
> 
> I think I'd mildly prefer to use the same strategy for getting the git repository as #159 did: i.e. just pull it down as part of the generation script. Otherwise it's a bit inconsistent about what we do with submodules. WDYT @marinelli ?

I'm not sure what the best strategy is, but I think it would be nice to:
- Reduce the number of tools required to build the package published on Hackage (eg: `git`, `make`, etc) 
- Allow a maintainer/developer to retrieve the code easily (like the protobuf proto files or the semantic conventions yaml files), and, if possible, generate the required Haskell modules that will be published in the package.

I wrote a POC for a shell script, it's based on the @kakkun61 branch, https://github.com/iand675/hs-opentelemetry/commit/1b98843eea38041038e0e4603ab320e4731c59c4.

marinelli (commented):
We have this `OpenTelemetry.SemanticConventions` template module.
I would move it outside the `src` dir (eg: in dev), than the `generate.hs` program will take the file, and place the generated module in the `src` folder.

Would it make sense to create different modules for different versions? Something like:
`OpenTelemetry.SemanticConventions.V1_24_0`, `OpenTelemetry.SemanticConventions.V1_25_0`, etc..

In this way we would disentangle the package version from the specification version. We could also have `OpenTelemetry.SemanticConventions.Latest` or simply `OpenTelemetry.SemanticConventions` that will re-export the module with the latest version.

marinelli:
> > Would it make sense to create different modules for different versions?
> 
> I guess a question would be: why are we not doing that for other packages we have that offer spec-based things. e.g. the `otlp` package?
> 
> Personally I would be surprised if people really care _so_ much about the precise version of the semantic conventions that they're getting (unless anyone already knows that they really care about that?) so I'm inclined to keep it simple?

I took inspiration by having a look at the code of a corresponding package for Go [1][2].
Anyway, I totally agree to maintain the code simple.

[1] https://github.com/open-telemetry/opentelemetry-go/tree/main/semconv
[2] https://pkg.go.dev/go.opentelemetry.io/otel/semconv

kakkun61:
Modules with semconv versions look nice, why not follow the official Go package's approach?

marinelli:
> Modules with semconv versions look nice, why not follow the official Go package's approach?

This could be introduced later. I do not think it's really necessary for now. We could just add a module like `OpenTelemetry.SemanticConventions`. This will contain the latest available version of the specification encoded as an Haskell module, and for now it could just contain the version 1.24.0 (I think the one you used in this PR), and we should also include a `version :: Data.Version.Version` value, and we should also add something similar in the OTLP module(s) (as @michaelpj suggested).

The goal of this PR, I think, should be to add this new package, and introduce a "resilient" way to update/introduce a new encoded version of the specifications.

kakkun61:
Yes, future feature if required!

kakkun61:
I request review again.

michaelpj (approved):
Okay, let's go with this for now!

PR #144, opened 2024-08-09
[http-client] using hooks

kakkun61:
# What

- to use `managerModifyRequest` and `managerModifyResponse` to implement this instrumentation
  - instead of wrapping original http-client's `Manager`

# Why

- a lot of wrapping and unwrapping are necessary because the wrapper type is different from the original one of course when using http-client instrumentation

# How

- `ManagerSetting` provides hooks for requests and responses
- to call `createSpan` in a hook for a request and to call `endSpan` in a hook for a response
- `ThreadLocalStorage` is used to transport span data created in a hook for a request to a hook for a response

# Note

Cachix @sandydoo seems to implement a similar instrumentation.

https://github.com/cachix/hs-opentelemetry-instrumentation-http-client/


kakkun61:
This is ready to be review!

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient.hs
```diff
@@ -13,143 +15,144 @@
...
+      , Orig.managerModifyResponse = managerModifyResponse >=> responseModifier tls
+      }
+
+
+type ThreadLocalStorage = TLS.TLS (IORef (Maybe Otel.Span))
+
+
+requestModifier :: HasCallStack => Otel.Tracer -> ThreadLocalStorage -> Orig.Request -> IO Orig.Request
+requestModifier tracer tls request = do
+  spanRef <- TLS.getTLS tls
```
michaelpj, line 69: So, there's some kind of important set of assumptions here about how requests and responses are handled. It looks to me like you're assuming that a response to a request will always be exactly the next response processed by the thread that sent the response. Is that true? It could do with some explanation as to why we think it is true if so!

(That assumption is important because it means we can use a thread-local storage to effectively identify a request: if a response comes back and there is something in the storage, it must be for the request that we sent.)

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient.hs
```diff
@@ -13,143 +15,144 @@
...
+
+
+responseModifier :: HasCallStack => ThreadLocalStorage -> Orig.Response Orig.BodyReader -> IO (Orig.Response Orig.BodyReader)
+responseModifier tls response = do
+  spanRef <- TLS.getTLS tls
+  maybeSpan <- readIORef spanRef
+  case maybeSpan of
+    Just span_ -> do
+      Otel.addAttributes span_ $ makeResponseAttributes response
+      Otel.endSpan span_ Nothing
```
michaelpj, line 90: What happens if there is no response? or if an exception is thrown? I think this version is less safe in the presence of failures than the old one. The old one used the `inSpan` functions, so should never leave us with unterminated spans, but I think this one can do?

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient.hs
```diff
@@ -13,143 +15,144 @@
...
+      writeIORef spanRef Nothing
+    Nothing -> assert False $ pure () -- something went wrong
+  pure response
+
+
+makeThreadLocalStorage :: IO ThreadLocalStorage
+makeThreadLocalStorage = TLS.mkTLS $ newIORef Nothing
+
+
+makeRequestAttributes :: Orig.Request -> Otel.AttributeMap
```
michaelpj, line 100: These changes look useful, though!

instrumentation/http-client/src/OpenTelemetry/Instrumentation/HttpClient.hs
kakkun61: That point is certainly very important. I had made that assumption unconsciously. I will check it out. Maybe I need to think of another way.

shinzui:
@kakkun61 @michaelpj I’m interested in using the instrumentation in this PR instead of what’s currently on master. Is there anything I can do to help move it forward?

PR #148, opened 2024-09-26
Support OTEL_SDK_DISABLED

michaelpj:
Very simply, we return the empty tracer provider options in this case, which makes tracing do nothing.


sdk/src/OpenTelemetry/Trace.hs
```diff
@@ -333,30 +335,35 @@ getTracerProviderInitializationOptions = getTracerProviderInitializationOptions'
...
-        emptyTracerProviderOptions
-          { tracerProviderOptionsIdGenerator = defaultIdGenerator
-          , tracerProviderOptionsSampler = sampler
-          , tracerProviderOptionsAttributeLimits = attrLimits
-          , tracerProviderOptionsSpanLimits = spanLimits
-          , tracerProviderOptionsPropagators = propagators
-          , tracerProviderOptionsResources = materializeResources allRs
-          }
-  pure (processors, providerOpts)
+  disabled <- fmap (== "true") <$> lookupEnv "OTEL_SDK_DISABLED"
```
marinelli, line 338: I think we should use a _case-insensitive_ comparison here (see https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value).
We should also fix the other occurrences of `fmap (== "true") <$> lookupEnv` in `OpenTelemetry.Exporter.OTLP.Span`.

michaelpj: Good point, I misread that as case-_insensitive_ :sweat_smile: 

I agree the rest are also all wrong. If it's okay, I'll make a PR on top of this one fixing them all, so we remain consistent.

marinelli (approved):

sdk/src/OpenTelemetry/Trace.hs
marinelli: I created a (draft) PR for this https://github.com/iand675/hs-opentelemetry/pull/153.

pbrisbin (approved):
LGTM, and agreed about fixing up the boolean handling. @marinelli if you mark that PR Ready for Review, I can review it.

marinelli:
> LGTM, and agreed about fixing up the boolean handling separately. @marinelli if you mark that PR Ready for Review, I can review it.

done

PR #149, opened 2024-10-02
Add instrumentation for tasty

michaelpj:
This adds spans for all of:
1. Individual test cases
2. Test groups
3. Resource setup and teardown

It also correctly handles test parallelism.


michaelpj:
@ocharles this is just the library that I wrote for us to use

ocharles (approved):

PR #150, opened 2024-10-24
Bump OTLP specification to version 1.0.0

marinelli:


otlp/package.yaml
```diff
@@ -1,7 +1,7 @@
 _common/lib: !include "../package-common.yaml"
 
 name:                hs-opentelemetry-otlp
-version:             0.0.1.0
+version:             0.0.2.0
```
michaelpj, line 4: This still needs to be reverted

michaelpj:
@marinelli are you familiar with the spec changes that have occurred? In particular, it looks like we should maybe be renaming our `InstrumentationLibrary` to `InstrumentationScope`? Do you know if there are any other things like this we might be missing?

otlp/package.yaml
marinelli: done

marinelli:
> Thanks, this looks good to me, modulo the versioning change. Have you tested that a third-party OTLP consumer still accepts the data we produce with this change?

I tested the library by sending tracing messages to _Grafana tempo_ (with and without an `opentelemetry-collector` in the middle).

michaelpj:
Great, thanks!

marinelli:
> @marinelli are you familiar with the spec changes that have occurred? In particular, it looks like we should maybe be renaming our `InstrumentationLibrary` to `InstrumentationScope`? Do you know if there are any other things like this we might be missing?

I had a look at the `opentelemetry-proto` [changelog](https://github.com/open-telemetry/opentelemetry-proto/blob/7312bdf63218acf27fe96430b7231de37fd091f2/CHANGELOG.md), at the [diff](https://github.com/open-telemetry/opentelemetry-proto/compare/v0.14.0...v1.0.0) between versions, and I didn't find any other mandatory changes relevant for the new OTLP version.

I think we could rename `InstrumentationLibrary` to `InstrumentationScope`, but I'm still not sure why we have `librarySchemaUrl` there :) (I'm still learning OpenTelemetry)

PR #151, opened 2024-10-28
Fix attribute length limits

### LINKED ISSUE #113: Attribute max length calculation uses the wrong option value
https://github.com/iand675/hs-opentelemetry/blob/02653bbfa286579be13cef17701da454ac3dfce1/api/src/OpenTelemetry/Attributes.hs#L92-L94

newAttrs should use `attributeLengthLimit `, otherwise attributes get cut at 128 bytes by default.

```
newAttrs = H.insert k (maybe id limitLengths attributeLengthLimit $ toAttribute v) attributes
```

michaelpj:
Fixes #113


sdk/test/Spec.hs
```diff
@@ -3,11 +3,14 @@ import qualified OpenTelemetry.ContextSpec as ContextSpec
...
 import OpenTelemetry.Trace (initializeGlobalTracerProvider)
 import qualified OpenTelemetry.TraceSpec as TraceSpec
+import System.Environment (setEnv)
 import Test.Hspec
 
 
 main :: IO ()
 main = do
+  -- For the attribute length limit test
+  setEnv "OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT" "50"
```
ocharles, line 13: Why doesn't the test itself set this? Do we need this action-at-a-distance?

sdk/test/OpenTelemetry/TraceSpec.hs
```diff
@@ -208,6 +209,17 @@ spec = describe "Trace" $ do
...
+    specify "Attribute length limit is respected" $ asIO $ do
+      p <- getGlobalTracerProvider
+      let t = makeTracer p "woo" tracerOptions
+      s <- createSpan t Context.empty "test_span" defaultSpanArguments
+      addAttribute s "attr1" ("looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong" :: Text)
+      addAttributes s (HM.singleton "attr2" (toAttribute @Text "looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong"))
+      s <- unsafeReadSpan s
+
+      lookupAttribute (spanAttributes s) "attr1" `shouldBe` Just (toAttribute @Text "looooooooooooooooooooooooooooooooooooooooooooooooo")
+      lookupAttribute (spanAttributes s) "attr2" `shouldBe` Just (toAttribute @Text "looooooooooooooooooooooooooooooooooooooooooooooooo")
```
ocharles, line 221: Mildly not obvious what this test is doing. Could you add comments explaining the input and expected output? I wonder if it would be clearer to state the expectation a bit more declaratively: the actual attribute has the prefix of the input and it's length is <= the limit.

ocharles (approved):

sdk/test/Spec.hs
michaelpj: It isn't currently easy to set this locally. It gets read when we set up the global tracer provider, which we do on the next line, so it's not clear that we can do it at any other time.

sdk/test/OpenTelemetry/TraceSpec.hs
michaelpj: Yeah, I made this simpler.

sdk/test/Spec.hs
ocharles: Ok let's add a comment both here, and where we need it

ocharles: Nvm I see you've done just that!

Issue #141, opened 2024-08-01
No way to create links conveniently


michaelpj:
The only way to do it that I can see is:
- Pass `[NewLink]` in when creating a span
- `NewLink` needs a `SpanContext`
- Use `getSpanContext` (not exported from the SDK!) to get the span context earlier

It would be nice to have something like 
```
-- makes a link to the given span
linkTo :: MonadIO m => Span -> Attributes -> m NewLink
-- adds the link to a span
addLink :: MonadIO m => Span -> NewLink -> m ()
```

RiugaBachi:
Hi all, has anyone taken a crack at this yet? I'm interested in making a PR for this as I've encountered similar limitations recently too.

ocharles:
I don't think anyone has started anything here yet @RiugaBachi - feel free to get stuck in!

PR #152, opened 2024-10-29
Make it easier to make links

michaelpj:
Fixes https://github.com/iand675/hs-opentelemetry/issues/141

- Add `addEvent`
    - This requires making the span link collection no longer frozen. At this point `FrozenBoundedCollection` is dead, so I dropped it.
- Export `getSpanContext` from the SDK.
    - This makes it easier to actually make `NewLink`s.

### DIFF
--- a/api/src/OpenTelemetry/Trace/Core.hs
+++ b/api/src/OpenTelemetry/Trace/Core.hs
@@ -94,8 +94,6 @@ module OpenTelemetry.Trace.Core (
   SpanKind (..),
   defaultSpanArguments,
   SpanArguments (..),
-  NewLink (..),
-  Link (..),
 
   -- ** Recording @Event@s
   Event (..),
@@ -111,6 +109,9 @@ module OpenTelemetry.Trace.Core (
   ToAttribute (..),
   PrimitiveAttribute (..),
   ToPrimitiveAttribute (..),
+  Link (..),
+  NewLink (..),
+  addLink,
 
   -- ** Recording error information
   recordException,
@@ -278,8 +279,8 @@ createSpanWithoutCallStack t ctxt n args@SpanArguments {..} = liftIO $ do
                           emptyAttributes
                           (H.unions [additionalInfo, attrs, attributes])
                     , spanLinks =
-                        let limitedLinks = fromMaybe 128 (linkCountLimit $ tracerProviderSpanLimits $ tracerProvider t)
-                        in frozenBoundedCollection limitedLinks $ fmap freezeLink links
+                        let emptyLinks = emptyAppendOnlyBoundedCollection $ fromMaybe 128 (linkCountLimit $ tracerProviderSpanLimits $ tracerProvider t)
+                        in foldl (\c l -> appendToBoundedCollection c l) emptyLinks (fmap (freezeLink t) links)
                     , spanEvents = emptyAppendOnlyBoundedCollection $ fromMaybe 128 (eventCountLimit $ tracerProviderSpanLimits $ tracerProvider t)
                     , spanStatus = Unset
                     , spanStart = st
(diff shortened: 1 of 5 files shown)


api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -505,6 +499,22 @@ addEvent (FrozenSpan _) _ = pure ()
 addEvent (Dropped _) _ = pure ()
 
 
+-- | Add a link to a recording span.
+addLink :: (MonadIO m) => Span -> NewLink -> m ()
+addLink (Span s) l = liftIO $ do
+  modifyIORef' s $ \(!i) -> i {spanLinks = appendToBoundedCollection (spanLinks i) (freezeLink (spanTracer i) l)}
+addLink (FrozenSpan _) _ = pure ()
+addLink (Dropped _) _ = pure ()
```
pbrisbin, line 507: I see this casing happens a lot throughout this module.

```console
% g grep 'Dropped.*pure ()$'
api/src/OpenTelemetry/Trace/Core.hs:addAttribute (Dropped _) _ _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:addAttributes (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:addEvent (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:setStatus (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:alterFlags (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:updateName (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:endSpan (Dropped _) _ = pure ()
api/src/OpenTelemetry/Trace/Core.hs:whenSpanIsRecording (Dropped _) _ = pure ()
```

I think a shared function like,

```hs
withSpan f = \case
  Span s -> f s
  FrozenSpan{} -> pure ()
  Dropped{} -> pure ()
```

Could be nice to have.

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -278,8 +279,8 @@ createSpanWithoutCallStack t ctxt n args@SpanArguments {..} = liftIO $ do
                           emptyAttributes
                           (H.unions [additionalInfo, attrs, attributes])
                     , spanLinks =
-                        let limitedLinks = fromMaybe 128 (linkCountLimit $ tracerProviderSpanLimits $ tracerProvider t)
-                        in frozenBoundedCollection limitedLinks $ fmap freezeLink links
+                        let emptyLinks = emptyAppendOnlyBoundedCollection $ fromMaybe 128 (linkCountLimit $ tracerProviderSpanLimits $ tracerProvider t)
+                        in foldl (\c l -> appendToBoundedCollection c l) emptyLinks (fmap (freezeLink t) links)
```
pbrisbin, line 283: Just want to make sure I understand: before, the collection itself and every link inside was frozen. Now the links inside are still each frozen, but the collection itself can be appended to. It makes sense you'd want to do this, but what are the ramifications, if any, of the collection no longer being frozen? Did it not need to be frozen for any particular reason originally?

pbrisbin (approved):

api/src/OpenTelemetry/Trace/Core.hs
iand675: This is due to the fact that v0.7.0 of the spec specified this behaviour:

> During the Span creation user MUST have the ability to record links to other Spans. Linked Spans can be from the same or a different trace. See [Links description](https://github.com/open-telemetry/opentelemetry-specification/blob/v0.7.0/specification/overview.md#links-between-spans). Links cannot be added after Span creation.

michaelpj: Aha, that makes a ton of sense. I had wondered why it was like that.

sdk/src/OpenTelemetry/Trace.hs
```diff
@@ -125,6 +126,7 @@ module OpenTelemetry.Trace (
   NewEvent (..),
   addEvent,
   inSpan'',
+  getSpanContext,
```
iand675, line 129: We already export `getSpanContext` from `OpenTelemetry.Trace.Core` in the API package. https://hackage.haskell.org/package/hs-opentelemetry-api-0.1.0.0/docs/OpenTelemetry-Trace-Core.html#v:getSpanContext

Is there a reason that we feel like we want it exported from the SDK here?

michaelpj: I guess I thought it was odd that we re-export _most_ functions for working with traces from `OpenTelemetry.Trace` here, but not this one. As it is, if you import `OpenTelemetry.Trace`, you get the `NewEvent` constructor but you can't do anything with it.

I guess more generally I'm unsure what our policy is for re-exports from `OpenTelemetry.Trace`. I was thinking along the lines of "everything you need to work with traces", which I think argues in favour of `getSpanContext`?

michaelpj: I've dropped this change to keep it uncontroversial

Issue #110, opened 2024-02-14
Improving project stewardship


iand675:
I've been feeling lately that I'm not necessarily responding to issues, PRs, and upstream evolutions of the OpenTelemetry spec in a timely fashion.

My desire for this project has been to build tracing that is fast, stable, relatively easy to use, and that supports most popular Haskell libraries. Ultimately, I'd like to contribute the library to the OpenTelemetry project itself. However, I've both had a new baby and had my oldest child pass away suddenly in the past year, so I have very little time or motivation to really work on anything outside of what is needed on this project to support the company that I work for.

So, I'd like to bring on one or two more people as core owners / contributors that will feel empowered to make decisions and shepherd the project into becoming mature enough that it becomes a go-to tool for production Haskell users. If you are interested, please let me know, and we can figure out next steps.

iand675:
Tagging a few people here that have previously contributed: @kakkun61 @solomon-b @ocharles @claymager

solomon-b:
Hi @iand675. I am so sorry for your loss I can't imagine what you are going through. 

I have the same desires for Haskell OpenTelemetry support. I don't think I have the bandwidth to do a lot of major feature work but I could help with code reviews, generating more regular maintenance releases, and the occasional feature work.

ocharles:
I'd like to echo @solomon-b - that sounds like an incredibly difficult situation, and I'd love to do what I can to help. You've done a great job with this library and we use it production at CircuitHub, so I'd like to give back. I'm not deeply involved with OpenTelemetry, but I'm happy to join as a maintainer to help keep things ticking along and building. Feel free to add me both here and on Hackage and I'll do what I can!

kakkun61:
Hello😊 I am working to make additional instrumentations at HERP's request. HERP is one of the companies that sponsor the Haskell Foundation. HERP is already using hs-opentelemetry on a production environment.

I would like to support this project.

velveteer:
Hey Ian, I am also willing to help maintain this library. We are using it at H-E-B and we really like it.

develop7:
First of all let me express late condolences and, yeah, I cannot possibly grasp the idea of what are you going through. I'm working on integrating OTel into PostgREST and would like to offer my help co-maintaining the project as well.

michaelpj:
[...]

PR #153, opened 2024-10-29
Fix reading boolean environment variables

marinelli:
As written in the specification [1], a boolean environment variable is true iff is equal to the case-insensitive string "true".

[1] https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value

Waiting for https://github.com/iand675/hs-opentelemetry/pull/148.

### DIFF
--- a/api/src/OpenTelemetry/Environment.hs
+++ b/api/src/OpenTelemetry/Environment.hs
@@ -13,5 +13,5 @@ isTrue :: String -> Bool
 isTrue = ("true" ==) . map C.toLower
 
 
-lookupBooleanEnv :: String -> IO (Maybe Bool)
-lookupBooleanEnv = fmap (fmap isTrue) . lookupEnv
+lookupBooleanEnv :: String -> IO Bool
+lookupBooleanEnv = fmap (maybe False isTrue) . lookupEnv
--- a/api/src/OpenTelemetry/Environment.hs
+++ b/api/src/OpenTelemetry/Environment.hs
@@ -1,12 +1,17 @@
 module OpenTelemetry.Environment (
-  isTrue,
+  lookupBooleanEnv,
 ) where
 
-import qualified Data.CaseInsensitive as CI
+import qualified Data.Char as C
+import System.Environment (lookupEnv)
 
 
--- Have a look here for the specification.
--- https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value
-
+{- | Does the given value of an environment variable correspond to "true" according
+to [the OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value)?
+-}
 isTrue :: String -> Bool
-isTrue = ("true" ==) . CI.mk
+isTrue = ("true" ==) . map C.toLower
+
+
+lookupBooleanEnv :: String -> IO (Maybe Bool)
+lookupBooleanEnv = fmap (fmap isTrue) . lookupEnv
--- a/api/src/OpenTelemetry/Environment.hs
+++ b/api/src/OpenTelemetry/Environment.hs
@@ -1,0 +1,12 @@
+module OpenTelemetry.Environment (
+  isTrue,
+) where
+
+import qualified Data.CaseInsensitive as CI
+
+
+-- Have a look here for the specification.
+-- https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value
+
+isTrue :: String -> Bool
+isTrue = ("true" ==) . CI.mk
--- a/exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Span.hs
+++ b/exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Span.hs
@@ -72,7 +72,7 @@ import Network.HTTP.Types.Header
 import Network.HTTP.Types.Status
 import OpenTelemetry.Attributes
 import qualified OpenTelemetry.Baggage as Baggage
-import qualified OpenTelemetry.Environment as Env
+import OpenTelemetry.Environment
 import OpenTelemetry.Exporter.Span
 import OpenTelemetry.Resource
 import OpenTelemetry.Trace.Core (timestampNanoseconds)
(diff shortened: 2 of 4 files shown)


api/src/OpenTelemetry/Environment.hs
```diff
@@ -0,0 +1,12 @@
...
+) where
+
+import qualified Data.CaseInsensitive as CI
+
+
+-- Have a look here for the specification.
+-- https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value
+
+isTrue :: String -> Bool
+isTrue = ("true" ==) . CI.mk
```
pbrisbin, line 12: In #110, Ian mentions minimal dependencies as a project goal. Should we just do `Data.Char` here instead of pulling in `case-insensitive`? If we're only comparing to `[t, r, u, e]`, I can't imagine there are any gotchas that can affect us.

```suggestion
isTrue = ("true" ==) . map Char.toLower
```

I would also consider exporting something like `lookupBooleanEnv = fmap (maybe False isTrue) . lookupEnv` from this module. Since you've named it `OpenTelementry.Environment`, it seems a reasonable expectation to me that it would own the actual env interaction, and not just the `String` predicate. WDYT?

marinelli: I totally agree about reducing dependencies. I choose `case-insensitive` 'cause it's already in the list of the dependencies in multiple packages.
I originally wrote something like `lookupBooleanEnv`, But I would like to propose some small refactoring for collecting all the environment variables (eg: we could use `getEnvironment` instead of using `lookupEnv` so many times). Anyway, I can add a function like that, it would make `loadExporterEnvironmentVariables` a bit more readable.

pbrisbin: > I choose case-insensitive 'cause it's already in the list of the dependencies in multiple packages

Ah, fair point!

iand675: To clarify my stance on dependencies a bit, the ones I want to keep as minimal as possible are specifically `hs-opentelemetry-api` and `hs-opentelemetry-sdk`. Especially `api`. The reasoning for that is that it's intended to be something that other library writers can use in their code as a dependency. It makes it far easier for them to upgrade their package bounds if there are fewer dependencies.

Integration packages I'm far less fussed about.

api/src/OpenTelemetry/Environment.hs
```diff
@@ -0,0 +1,12 @@
...
+  isTrue,
+) where
+
+import qualified Data.CaseInsensitive as CI
+
+
+-- Have a look here for the specification.
+-- https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value
+
+isTrue :: String -> Bool
```
michaelpj, line 11:
```suggestion
-- | Does the given value of an environment variable correspond to "true" according
-- to [the OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value)?
isTrue :: String -> Bool
```

api/src/OpenTelemetry/Environment.hs
michaelpj: I agree that in this instance I think we can go with the dumb version.

api/src/OpenTelemetry/Environment.hs
marinelli: done

api/src/OpenTelemetry/Environment.hs
marinelli: done

api/src/OpenTelemetry/Environment.hs
```diff
@@ -0,0 +1,17 @@
...
+
+{- | Does the given value of an environment variable correspond to "true" according
+to [the OpenTelemetry specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#boolean-value)?
+-}
+isTrue :: String -> Bool
+isTrue = ("true" ==) . map C.toLower
+
+
+lookupBooleanEnv :: String -> IO (Maybe Bool)
+lookupBooleanEnv = fmap (fmap isTrue) . lookupEnv
```
pbrisbin, line 17: > Any value not explicitly defined here as a true value, **including unset** and empty values, MUST be interpreted as false

I think it would be better to handle all aspects of the spec here, rather than push a `Nothing` to the callers (to potentially be misinterpreted).


```suggestion
lookupBooleanEnv :: String -> IO Bool
lookupBooleanEnv = fmap (maybe False isTrue) . lookupEnv
```

marinelli: From spec: _Any value not explicitly defined here as a true value, including unset and empty values, MUST be interpreted as false._
I see :)

marinelli: done

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Span.hs
```diff
@@ -108,9 +109,9 @@ data OTLPExporterConfig = OTLPExporterConfig
   { otlpEndpoint :: Maybe String
   , otlpTracesEndpoint :: Maybe String
   , otlpMetricsEndpoint :: Maybe String
-  , otlpInsecure :: Maybe Bool
-  , otlpSpanInsecure :: Maybe Bool
-  , otlpMetricInsecure :: Maybe Bool
+  , otlpInsecure :: Bool
+  , otlpSpanInsecure :: Bool
+  , otlpMetricInsecure :: Bool
```
pbrisbin, line 114: Surely these are used by something that'll break now, right? :thinking:

pbrisbin (approved):
Code LGTM, except I'm not sure what the removal of the `Maybe`s will affect. Also, did you want to use this for `OTEL_SDK_DISABLE` too?

michaelpj:
> Also, did you want to use this for OTEL_SDK_DISABLE too?

Haven't merged that yet!

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Span.hs
michaelpj: CI passes, so I guess not. It seems they're just parsed and not used.

api/src/OpenTelemetry/Environment.hs
michaelpj: Good catch!

michaelpj (approved):

PR #156, opened 2024-11-12
Fix resource merge

### LINKED ISSUE #111: Resource Merge Prioritizes Old Resource
The function `mergeResources` in `OpenTelemetry.Resource` seems to be backwards. The documentation calls the first argument the old resource and the second argument the updating resource, meaning the second argument's attribute values should take precedence when a key is in both resources. However, due to the underlying `HashMap`'s left-bias, the old value is taking precedence. For example:

```haskell
l = mkResource ["host.name" .= ("left" :: Text)] :: Resource 'Nothing
r = mkResource ["host.name" .= ("right" :: Text)] :: Resource 'Nothing
getMaterializedResourcesAttributes . materializeResources $ mergeResources l r
-- Attributes {attributes = fromList [("host.name",AttributeValue (TextAttribute "left"))], attributesCount = 2, attributesDropped = 0}
```

As a result, `getTracerProviderInitializationOptions'` does not prioritize the user-provided resource attributes. 
This makes it very difficult to override the built-in resources (such as the host detector).

Similarly, I think `getTracerProviderInitializationOptions'` should prioritize the attributes read from `OTEL_RESOURCE_ATTRIBUTES` above the detected resources. Currently we have:

```haskell
let allRs = mergeResources (builtInRs <> envVarRs) rs
```

Since `<>` is defined in a similar way as `mergeResources` and is also left-biased, both issues could be resolved by updating `<>` and `mergeResources` to be right-biased. Alternatively, the documentation for `mergeResources` could be updated to reflect the current behavior, and `getTracerProviderInitializationOptions'` could be updated to to reverse the order of arguments.

ehofreiter:
Fixes #111

- Re-order the Resource merge in `getTracerProviderInitializationOptions'` so that built-in resource attributes are _lowest_ priority rather than _highest_.
  - Enables overriding built-in resources with OTEL_RESOURCE_ATTRIBUTES environment variable.
  - Enables overriding OTEL_RESOURCE_ATTRIBUTES and built-ins with the user resource provided as an argument to `getTracerProviderInitializationOptions'`.
- Add unit tests and documentation to establish the left-biased nature of `Attributes` and `Resource`: when two attribute keys conflict, the value of the leftmost element is used.
- Fix bug in `unsafeMergeAttributesIgnoringLimits` where attribute count can get out of sync with actual size of the HashMap. Instead of adding the sizes of the inputs, simply take the size of the resulting HashMap. This is the approach used in the neighboring function `unsafeAttributesFromListIgnoringLimits`. Taking the size is $O(n)$ for HashMap, so may want to use Data.Map instead.

### DIFF
--- a/api/hs-opentelemetry-api.cabal
+++ b/api/hs-opentelemetry-api.cabal
@@ -110,8 +110,10 @@ test-suite hs-opentelemetry-api-test
   type: exitcode-stdio-1.0
   main-is: Spec.hs
   other-modules:
+      OpenTelemetry.AttributesSpec
       OpenTelemetry.BaggageSpec
       OpenTelemetry.Logs.CoreSpec
+      OpenTelemetry.ResourceSpec
       OpenTelemetry.SemanticsConfigSpec
       OpenTelemetry.Trace.SamplerSpec
       OpenTelemetry.Trace.TraceFlagsSpec
--- a/api/src/OpenTelemetry/Attributes.hs
+++ b/api/src/OpenTelemetry/Attributes.hs
@@ -260,8 +260,13 @@ instance (ToPrimitiveAttribute a) => ToAttribute [a] where
   toAttribute = AttributeArray . map toPrimitiveAttribute
 
 
+-- | Left-biased merge.
 unsafeMergeAttributesIgnoringLimits :: Attributes -> Attributes -> Attributes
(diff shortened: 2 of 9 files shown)


michaelpj (approved):
Looks great, thanks for this! Very much appreciate getting some more tests!

michaelpj:
It would be super to get some tests for the attribute limiting behaviour, as I think we've had other bugs there before, but this is great as is.

PR #157, opened 2024-11-13
[api] attribute map and type-safe attribute

kakkun61:
# Why is this

This PR is to introduce type-safer attributes.

# What's changed

New types `AttributeMap` and `Key` are come.

- attribute map is just `type AttributeMap = H.HashMap Text Attribute`
  - *without* counts and limits
- attribute key is `newtype Key a = Key {unkey :: Text}`
  - an key text of maps with a Haskell type corresponding to an attribute's type
  - automatic casting from/to Haskell type
- attributes is not changed
  - *with* counts and limits


api/src/OpenTelemetry/Attributes/Key.hs
```diff
@@ -0,0 +1,35 @@
...
+) where
+
+import Data.String (IsString (..))
+import Data.Text (Text)
+import qualified Data.Text as T
+import GHC.Generics (Generic)
+import OpenTelemetry.Attributes.Attribute (Attribute)
+
+
+newtype Key a = Key {unkey :: Text} deriving stock (Show, Eq, Ord, Generic)
```
michaelpj, line 25:
```suggestion
-- | A 'Key' is a name for an attribute. The type parameter sets the type of the
-- attribute the key should be associated with. This is useful for standardising 
-- attribute keys, since we can define both the key and the type of value it is 
-- intended to record.
--
-- For example, we might define:
-- @  
--     -- See https://opentelemetry.io/docs/specs/semconv/attributes-registry/server/
--     serverPortKey :: Key Int
--     serverPortKey = "server.port"
```

michaelpj: Having written this, I wonder if we shouldn't just call this `AttributeName`. It's more verbose, but perhaps better to be more descriptive?

api/src/OpenTelemetry/Attributes/Map.hs
```diff
@@ -0,0 +1,60 @@
...
+import OpenTelemetry.Attributes.Key (
+  Key (Key),
+ )
+import Prelude hiding (lookup, map)
+
+
+type AttributeMap = H.HashMap Text Attribute
+
+
+insertByKey :: ToAttribute a => Key a -> a -> AttributeMap -> AttributeMap
```
michaelpj, line 47: if we had called it `AttributeName`, then this would be `insertByName`, which I think sounds nice?

api/src/OpenTelemetry/Attributes.hs
```diff
@@ -70,7 +80,7 @@ defaultAttributeLimits =
 
 
 data Attributes = Attributes
```
michaelpj, line 82: It's a shame we're already using `Attributes` as the name here, since it would arguably be a better name for what you're calling `AttributeMap`.

michaelpj (approved):
I think this looks okay. It's mostly backwards-compatible apart from naming changes to some lesser-used functions, so I think it's not too disruptive.

api/src/OpenTelemetry/Attributes/Map.hs
ehofreiter: Another option: remove `ByKey` suffix and just call it `insert`.

michaelpj:
@ehofreiter if we were starting fresh I'd probably go with a scheme like this:

- Attribute: key-value pair
- AttributeKey: key of an attribute
- AttributeValue: value of an attribute, sometimes referred to simply as an "attribute" where the key is clear from context. So `ToAttribute` is I think still clear enough, but `ToAttributeValue` might be better.
- Attributes: a set of attributes (i.e. a set of key-value pairs i.e. a map)
- BoundedAttributes: a set of attributes with limits (mirrors what we do elsewhere with e.g. `BoundedCollection`)

But that would be a lot of churn, unclear if it's worth the benefit.

api/src/OpenTelemetry/Attributes/Key.hs
michaelpj: I think the general naming convention in the package has been to go for longer names that are non-ambiguous by themselves. e.g. we have `SpanProcessor` rather than just `Span.Processor`. So I think "AttributeKey" fits that approach better.

api/src/OpenTelemetry/Attributes.hs
iand675: Given that there are uniqueness constraints on attributes, I'm leaning towards AttributeMap somewhat so that the semantics are more obvious to consumers of the library.

> Some other implementations may use a streaming approach where every SetAttribute API call immediately results in that individual attribute value being exported using a streaming wire protocol. In such cases the enforcement of uniqueness will likely be the responsibility of the recipient of this data.

This bit of the spec is interesting though. I am trying to think of if there is any real benefit to that approach / whether that influences naming here in any way if we wanted to support that... but no issues come to mind.

api/src/OpenTelemetry/Attributes/Key.hs
kakkun61: Thanks for checking the spec for “key” vs. “name”. I prefer “key” for maps. I follow this project's customs although I prefer non-ambiguous name.

So I will rename this to `AttributeKey`.

api/src/OpenTelemetry/Attributes/Map.hs
kakkun61: Actually I want to rename this to `insert` and to remove the original `insert`. I kept `insert` because of compatibility.

api/src/OpenTelemetry/Attributes.hs
kakkun61: Maybe it is not necessary that `Attributes` has `attributesCount` and `attributesDropped` and another has, but I must read the spec more.

api/src/OpenTelemetry/Attributes/Attribute.hs
```diff
@@ -0,0 +1,224 @@
...
+class FromPrimitiveAttribute a where
+  fromPrimitiveAttribute :: PrimitiveAttribute -> Maybe a
+
+
+{- | An attribute represents user-provided metadata about a span, link, or event.
+
+ Telemetry tools may use this data to support high-cardinality querying, visualization
+ in waterfall diagrams, trace sampling decisions, and more.
+-}
+data Attribute
```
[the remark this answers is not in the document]
kakkun61: About `Attribute` and `AttributeValue`, we'll probably need to adjust them to the specifications later.

ehofreiter:
> @ehofreiter if we were starting fresh I'd probably go with a scheme like this:
> 
> * Attribute: key-value pair
> * AttributeKey: key of an attribute
> * AttributeValue: value of an attribute, sometimes referred to simply as an "attribute" where the key is clear from context. So `ToAttribute` is I think still clear enough, but `ToAttributeValue` might be better.
> * Attributes: a set of attributes (i.e. a set of key-value pairs i.e. a map)
> * BoundedAttributes: a set of attributes with limits (mirrors what we do elsewhere with e.g. `BoundedCollection`)
> 
> But that would be a lot of churn, unclear if it's worth the benefit.

@michaelpj: I like the naming scheme you laid out above, though it would indeed be a lot of churn. Probably not worth it for the names alone.

For now, would it be worth noting these slight differences with the spec in the Haskell doc comments? The docs reference and quote the spec in various places, so it could help clarify them. It could also provide a cue to consider renaming if the interface ever has to change for other reasons.

Issue #158, opened 2024-11-13
Avoid build-time code generation


michaelpj:
I think we should seriously consider avoiding any build-time code generation. At the moment that's mostly `hs-opentelemetry-otlp`, but there is also https://github.com/iand675/hs-opentelemetry/pull/142 which I think is desirable.

Why?

1. It requires custom setups. Custom setups are generally harder for tooling to deal with, cost you parallelism, etc. The general feeling of the cabal developers is that they wish they didn't exist :joy: 
2. It puts more requirements onto our users. Notably, currently users need to have `protoc` around in order to build `hs-opentelemetry-otlp`. Cabal does not know this, it is quite opaque to users.

There is an easy way to avoid this: just do the codegen manually and check in the result. It's ugly, and involves checking in generated code, but it's robust and not too painful if the input doesn't change that often (which I don't think it does).

I also note that the last release of `hs-opentelemetry-otlp` did _not_ use a custom setup, so this is going to be a new breaking requirement for our users. We [already](https://github.com/iand675/hs-opentelemetry/issues/154#issue-2633250118) have people complaining about that.

Overall, I would prefer to revert https://github.com/iand675/hs-opentelemetry/pull/122 before we release, and ideally commit to not doing this in the future.

cc @iand675 and @evanlauer1 , who are likely to have opinions.

michaelpj:
@marinelli has a PR up reverting to checked in generated code. I'll leave it a while in case anyone has any objections, otherwise I'll merge that.

develop7:
> There is an easy way to avoid this: just do the codegen manually and check in the result. It's ugly, and involves checking in generated code, but it's robust and not too painful if the input doesn't change that often (which I don't think it does).

Did a quick glance at https://github.com/open-telemetry/opentelemetry-proto/commits/main/ and it gets updated every few months or so, which doesn't seem often enough to warrant a custom setup.

michaelpj:
We're back to manual generation for now.

PR #159, opened 2024-11-15
Add a shell script for generating OTLP modules

marinelli:
This PR proposes to use a shell script for generating the Haskell OTLP modules.
It should fix https://github.com/iand675/hs-opentelemetry/issues/158.


otlp/scripts/generate-modules.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
michaelpj, line 1: Please update the README in this directory to explain how to use this.

otlp/scripts/generate-modules.sh
```diff
@@ -0,0 +1,69 @@
...
+
+set -eo pipefail
```
michaelpj, line 3: I'd have been tempted to write this in Haskell as an executable in the package, but this is also good :)

otlp/scripts/generate-modules.sh
```diff
@@ -0,0 +1,69 @@
...
+fi
+
+if ! command -v proto-lens-protoc >/dev/null 2>&1; then
+  printf '%s\n' 'proto-lens-protoc is not available' 1>&2
+  exit 1
+fi
+
+PROTO_LENS=$(command -v proto-lens-protoc)
+
+OTLP_VERSION="$1"
```
michaelpj, line 17: I think we should write this down in a file. Could just be a constant in this script.
1. That makes it easy to see what version we support without having to check the git history.
2. It lets us run a CI check to assert that the generated files haven't changed from what we would get if we re-generated them, which guards us against accidental modification (although that probably won't happen, and I don't think we should block this PR on having such a check).

otlp/src/Proto/Opentelemetry/Proto/Collector/Logs/V1/LogsService.hs
```diff
@@ -0,0 +1,641 @@
+{- HLINT ignore -}
```
michaelpj, line 1: I suggest telling github that these files are generated: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

otlp/scripts/generate-modules.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+if ! command -v protoc >/dev/null 2>&1; then
+  printf '%s\n' 'protoc is not available' 1>&2
+  exit 1
+fi
+
+if ! command -v proto-lens-protoc >/dev/null 2>&1; then
```
michaelpj, line 10: For us nix users, can you add `haskellPackages.proto-lens-protoc` here: https://github.com/iand675/hs-opentelemetry/blob/main/flake.nix#L98

We've already got `protoc`, so that's fine.

otlp/scripts/generate-modules.sh
michaelpj: I guess we could even write it into a Haskell module and export it as a value from the library, although I'm not sure if there's actually any use for that!

michaelpj (approved):
Great, thanks for doing this! I was just about to look at it :joy:

marinelli:
> Great, thanks for doing this! I was just about to look at it :joy: 

I reused/refactored something I wrote for another OpenTelemetry Haskell library [1] 😄

[1] https://github.com/Siprj/otel-haskell-sdk

otlp/scripts/generate-modules.sh
marinelli: I created the `OTLP_VERSION` file.

otlp/src/Proto/Opentelemetry/Proto/Collector/Logs/V1/LogsService.hs
marinelli: done

otlp/scripts/generate-modules.sh
marinelli: done

otlp/scripts/generate-modules.sh
marinelli: Done, and feel free to update it :)

michaelpj:
Great, I'm happy, I'm going to leave it a few days in case there is any dissent, and then merge.

develop7:
Beautiful! :shipit:

otlp/.gitattributes
```diff
@@ -1 +1,2 @@
+proto/**/*.proto linguist-generated=true
```
marinelli, line 1: @michaelpj I added this line to hide the diffs for the protobuf files too.
You can have a look here https://github.com/marinelli/hs-opentelemetry/pull/1, just to be sure it works as expected.

PR #161, opened 2024-11-19
Add a TH helper for creating InstrumentationLibrarys

### LINKED ISSUE #2: Figure out how to get library versions without hardcoding them

michaelpj:
This adds a little TH helper that parses out the library name and version from the TH location, so you can reliably get the right answer.

Fixes #2


ocharles (approved):
I didn't look closely but I guess this is just what we wrote at CircuitHub?

michaelpj:
Yes, exactly.

PR #172, opened 2024-12-17
Improve source code attribute handling

michaelpj:
My previous attempt at this was wrong for a few reasons.
1. I got the locations misaligned within the call stack, so we were giving the wrong call location.
2. It's in fact just totally unclear what to do with frozen call stacks in general.

This commit fixes 1 and enables a different approach to 2: get the user to explicitly supply source code attributes if they are doing something clever. So rather than trying to freeze the call stack to get us to do waht you want, you can just create the source code attributes yourself (using `callerAttributes` if you like) and add them. The main change to enable this is just to not add source code attributes from the call stack if there are any already present, which prevents us getting a muddled mix of attributs if only some have been set.

### DIFF
--- a/api/src/OpenTelemetry/Trace/Core.hs
+++ b/api/src/OpenTelemetry/Trace/Core.hs
@@ -209,7 +209,9 @@ createSpan
   -- ^ Additional span information
   -> m Span
   -- ^ The created span.
-createSpan t ctxt n args = createSpanWithoutCallStack t ctxt n (args {attributes = H.union (attributes args) callerAttributes})
+  -- Try and infer source code information unless the user has set any of the attributes already, which
+  -- we take as an indication that our automatic strategy won't work well.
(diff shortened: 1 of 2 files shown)


develop7:
Does this enable an user to declare an `inSpan` convenience wrapper and, with healthy amount of fiddling with `CallStack`, have correct code locations reported instead of ones inside the wrapper?

michaelpj:
Yes, that's also what I want to do. Previously I thought what I had done would allow you to do this using `popCallStack`, but it doesn't work properly for various reasons :( 

With this version you can do it; you just have to set the source attributes yourself. I should add an example as a test! This is moderately annoying, but I think overall it's better to say "control the attributes by setting them directly" rather than by saying "try to manipulate the call stack so it matches what the library expects".

develop7:
Yeah, the example would be especially handy

kakkun61 (approved):

michaelpj:
Added an example. It's fairly rational really: your wrapper now just needs to do pretty much exactly what the span creation functions in `OpenTelemetry.Trace.Core` do.

PR #174, opened 2025-01-09
Export `toImmutableSpan`

turnage:
At the moment, the thread-local storage retains a span, but the `Span` exposed there cannot be traversed upward. This only becomes available in the `Processor`s.

We have a use case that would benefit from being able to traverse up the `Span` with `alterSpansUpward` before the span is concluded and sent to the `Processor`.

The use case is for a span to indicate that not only itself, but also all child spans, should not propagate span information over an async boundary (in an application that uses this library). We want to use a span attribute for this (it's useful as an attribute both for this control behavior and as an indicator in our telemetry viewer).

### DIFF
--- a/api/ChangeLog.md
+++ b/api/ChangeLog.md
@@ -1,5 +1,9 @@
 # Changelog for hs-opentelemetry-api
 
+## 0.2.1.0
+
+- defined and exported `toImmutableSpan` and `FrozenOrDropped` from `OpenTelemetry.Trace.Core`
+
 ## 0.2.0.0
 
 - `callerAttributes` and `ownCodeAttributes` now work properly if the call stack has been frozen. Hence most
--- a/api/src/OpenTelemetry/Internal/Trace/Types.hs
+++ b/api/src/OpenTelemetry/Internal/Trace/Types.hs
@@ -306,6 +306,14 @@ instance Show Span where
   showsPrec d (Dropped ctx) = showParen (d > 10) $ showString "Dropped " . showsPrec 11 ctx
 
 
+-- | Extracts the values from a @Span@ if it is still mutable. Returns @Nothing@ if the @Span@ is frozen or dropped.
+toImmutableSpan :: Span -> IO (Maybe ImmutableSpan)
+toImmutableSpan s = case s of
+  Span ioref -> Just <$> readIORef ioref
+  FrozenSpan ctx -> pure Nothing
+  Dropped ctx -> pure Nothing
+
+
 {- | TraceFlags with the @sampled@ flag not set. This means that it is up to the
  sampling configuration to decide whether or not to sample the trace.
 -}
--- a/api/src/OpenTelemetry/Internal/Trace/Types.hs
+++ b/api/src/OpenTelemetry/Internal/Trace/Types.hs
@@ -306,12 +306,15 @@ instance Show Span where
   showsPrec d (Dropped ctx) = showParen (d > 10) $ showString "Dropped " . showsPrec 11 ctx
 
 
+data FrozenOrDropped = SpanFrozen | SpanDropped deriving (Show, Eq)
+
+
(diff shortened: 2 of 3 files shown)


api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -72,6 +72,8 @@ module OpenTelemetry.Trace.Core (
 
   -- * Span operations
   Span,
+  toImmutableSpan,
+  FrozenOrDropped,
```
iand675, line 76: I think we need to export constructors here, but otherwise LGTM

iand675 (changes requested):

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -688,10 +690,12 @@ spanIsRemote (Dropped _) = pure False
...
-unsafeReadSpan = \case
-  Span ref -> liftIO $ readIORef ref
-  FrozenSpan _s -> error "This span is from another process"
-  Dropped _s -> error "This span was dropped"
+unsafeReadSpan s =
+  toImmutableSpan s >>= \case
+    Right span -> pure span
+    Left frozenOrDropped -> case frozenOrDropped of
+      SpanFrozen -> error "This span is from another process"
+      SpanDropped -> error "This span was dropped"
```
jkachmar, line 698: just outta curiosity, why the `error` call vs throwing an exception?

is the intention to preserve the thunk so that a frozen/dropped span doesn't immediately halt a test unless it's inspected?

iand675: Honestly, it's just that the code here has existed since the beginning and never been an issue. It would probably make sense to throw here, but it's a relatively niche function that has t needed much changing.

api/src/OpenTelemetry/Internal/Trace/Types.hs
```diff
@@ -306,6 +306,17 @@ instance Show Span where
   showsPrec d (Dropped ctx) = showParen (d > 10) $ showString "Dropped " . showsPrec 11 ctx
 
 
+data FrozenOrDropped = SpanFrozen | SpanDropped deriving (Show, Eq)
+
+
+-- | Extracts the values from a @Span@ if it is still mutable. Returns @Nothing@ if the @Span@ is frozen or dropped.
```
jkachmar, line 312:
```suggestion
-- | Extracts the values from a @Span@ if it is still mutable. Returns @Either@
-- of @SpanFrozen@ if the @Span@ is frozen or @SpanDropped@ if it is dropped.
```
ymmv on wording, but it looks like the `@Nothing@` doc might be a holdover from a prior revision

api/ChangeLog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for hs-opentelemetry-api
 
+## 0.2.1.0
+
+- defined and exported `toImmutableSpan` and `FrozenOrDropped` from `OpenTelemetry.Trace.Core`
+
```
jkachmar, line 6: nit: this feels like a patch more than a minor version bump

iand675 (approved):

api/ChangeLog.md
turnage: I'm following this guide

> Non-breaking change. Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were added to the interface, then A.B MAY remain the same but the new C MUST be greater than the old C. Note that modifying imports or depending on a newer version of another package may cause extra non-orphan instances to be exported and thus force a minor version change.

https://pvp.haskell.org/

Idk if that's the scheme adhered by this project though.

jkachmar: the PVP is weird in that it's `MAJOR.MAJOR.MINOR.PATCH` where you can omit the leading `MAJOR` if you choose to do so.

the current version is `0.2.0.0`, so `0.2.1.0` is a minor bump, whereas in the example on that page `0.2.0` -> `0.2.1` would be a patch bump.

PR #176, opened 2025-02-13
Implement SQLCommenter support

iand675:
This adds support for the [`sqlcommenter`](https://google.github.io/sqlcommenter/) spec, as outlined by folks at Google. We're using it at https://github.com/MercuryTechnologies for improved correlation between SQL queries and the rest of the OTel ecosystem. The main area we're seeing benefits is via https://pganalyze.com/, which has native support for the format. Namely, we can have "explain analyze" traces added to our existing application traces to help us understand why certain queries behave badly in production.

We've used this in prod for about 10 months, so I'm not particularly concerned about the efficacy of the implementation, but I do want to ensure that it builds against API changes & is updated on a similar cadence to other packages in this repo.

Why should this package live in this repo? Because Google donated it to the OpenTelemetry community: https://cloud.google.com/blog/products/databases/sqlcommenter-merges-with-opentelemetry


utils/sqlcommenter/src/SqlCommenter.hs
```diff
@@ -0,0 +1,268 @@
...
+  [] -> []
+  (x:xs) -> x : prependToAll sep xs where
+    prependToAll sep = \case
+      [] -> []
+      (x:xs) -> sep : x : prependToAll sep xs
+{-# INLINE intersperse #-}
+
+intercalate :: (Monoid a, Foldable f) => a -> f a -> a
+intercalate delim l = mconcat (intersperse delim l)
+{-# INLINE intercalate #-}
```
iand675, line 133: These are foldable versions of the list equivalent.

utils/sqlcommenter/src/SqlCommenter.hs
```diff
@@ -0,0 +1,268 @@
...
+    (0, pos) -> testBit w0 $ fromIntegral pos
+    (1, pos) -> testBit w1 $ fromIntegral pos
+    (2, pos) -> testBit w2 $ fromIntegral pos
+    (3, pos) -> testBit w3 $ fromIntegral pos
+    _        -> error "Word8 value out of bounds"
+
+-- SQL Commenter spec wants them escaped with a slash, but this should
+-- probably solve the same issue
+unreservedQS :: Word8Set
+unreservedQS = foldr insert SqlCommenter.empty $ map c2w "-_.~'"
```
iand675, line 120: This builds a data structure that can perform fast checks to determine whether to escape a character or not for a URL. This is similar to @ekmett's charset package, but with a much more constrained purposed, and is intended to avoid incurring extra dependencies. https://hackage.haskell.org/package/charset-0.3.11/docs/Data-CharSet.html

utils/sqlcommenter/src/SqlCommenter.hs
```diff
@@ -0,0 +1,268 @@
...
+sqlCommenterKey = unsafePerformIO $ Ctxt.newKey "sqlcommenter-attributes"
+{-# NOINLINE sqlCommenterKey #-}
+
+lookupSqlCommenterAttributes :: Ctxt.Context -> Map Text Text
+lookupSqlCommenterAttributes ctxt = case Ctxt.lookup sqlCommenterKey ctxt of
+  Nothing -> mempty
+  Just attrs -> attrs
+
+getSqlCommenterAttributes :: IO (Map Text Text)
+getSqlCommenterAttributes = lookupSqlCommenterAttributes <$> TL.getContext
```
jkachmar, line 54: I guess `sqlCommenterKey` has to be a global variable b/c of how `Vault` works, and since it's an implementation detail that never leaks beyond the scope of this module there shouldn't be a problem here?

checking my understanding: we can't make `sqlCommenterKey :: IO _` because we need `lookupSqlCommenterAttributes` to have a pure API for any callers who might need to use it in a pure context (even though `getSqlCommenterAttributes :: IO _`).

utils/sqlcommenter/src/SqlCommenter.hs
```diff
@@ -0,0 +1,268 @@
...
+
+addTraceDataToAttributes :: Core.Span -> Map Text Text -> IO (Map Text Text)
+addTraceDataToAttributes span attrs = do
+  (traceparent, tracestate) <- encodeSpanContext span
+  pure $ attrs
+    & M.insert "traceparent" (unsafeConvert traceparent)
+    & M.insert "tracestate" (unsafeConvert tracestate)
+  where
+    unsafeConvert = B.toText . B.unsafeFromByteString
```
jkachmar, line 63: [encodeSpanContext](https://hackage.haskell.org/package/hs-opentelemetry-propagator-w3c-0.0.1.4/docs/OpenTelemetry-Propagator-W3CTraceContext.html#v:encodeSpanContext) gives us two `ByteString`s.

it looks like these are produced by [traceparentHeader](https://hackage.haskell.org/package/hs-opentelemetry-propagator-w3c-0.0.1.4/docs/src/OpenTelemetry.Propagator.W3CTraceContext.html#local-6989586621679050537), which appears to construct these directly using a combination of primitives (`char7 '-'`, `word8HexFixed` on the `Word8` stored in `TraceFlags`) & `Base16`-encoded `SpanId` & `TraceId` values.

You've got tests for this, so presumably something upholds the invariant that all of these must be UTF-8 encoded text but I'm having trouble understand what exactly this is (if it is anything besides convention).

utils/sqlcommenter/test/Spec.hs
```diff
@@ -0,0 +1,52 @@
...
+main = hspec $ do
+  specify "Does not add comment if the query already has a comment" $ do
+    let queries =
+          [ "SELECT * FROM table -- noodle"
+          , "SELECT * FROM table -- noodle\n"
+          , "SELECT * FROM table /* noodle\n  poodle\n*/"
+          ]
+        someAttrs = M.fromList [("foo", "bar")]
+    for_ queries $ \query -> do
+      sqlCommenter query someAttrs `shouldBe` query
```
jkachmar, line 28: I don't quite follow what this is testing; `someAttrs` is just ignored entirely and the query is returned as-is because there is an existing comment?

utils/sqlcommenter/test/Spec.hs
```diff
@@ -0,0 +1,52 @@
...
+  specify "Parsing a queries reads attributes from the first comment" $ do
+    let query1 = "SELECT * FROM table -- foo='bar'\n-- bar='baz'"
+        query2 = "SELECT * FROM table /* noodle='poodle',wibble='wobble'*/ /* foo='bar' */"
+    parseFirstSqlComment query1 `shouldBe` M.fromList [("foo", "bar")]
+    parseFirstSqlComment query2 `shouldBe` M.fromList [("noodle", "poodle"), ("wibble", "wobble")]
+  specify "Parsing a query decodes encoded bytes" $ do
+    hedgehog $ do
+      let attrList = M.fromList <$> Gen.list (Range.linear 0 30) ((,) <$> textValGen <*> textValGen)
+      kvs <- forAll attrList
+      kvs === parseFirstSqlComment (sqlCommenter "" kvs)
```
jkachmar, line 52: I don't _think_ any of these tests cover the case where `sqlCommenter` actually modifies the query, right?

I think it'd be useful to add an end-to-end to check that span attributes get picked up tacked onto a query in comments according to the [sqlcommenter example documentation](https://google.github.io/sqlcommenter/) (or something).

utils/sqlcommenter/test/Spec.hs
iand675: Yes, exactly.

Per the sqlcommenter spec:

> If a comment already exists within a SQL statement, we MUST NOT mutate that statement.

utils/sqlcommenter/src/SqlCommenter.hs
iand675: Good question. The W3C `traceparent` and `tracestate` headers [by definition](https://www.w3.org/TR/trace-context/#traceparent-header-field-values) are US-ASCII encoded, so we don't have to think super hard about this assuming that `import OpenTelemetry.Propagator.W3CTraceContext (encodeSpanContext)` is implemented correctly.

utils/sqlcommenter/src/SqlCommenter.hs
iand675: Right. The main thing is that `sqlCommenterKey` needs to be _stable_ across uses. If we had users intitialize the key and pass it in any time they want to use it to look up the sqlcommenter attributes, then we could make it be an IO value. However, I don't think that's a great experience, and is relatively error prone compared to just making sure that there's a stable reference available.

utils/sqlcommenter/test/Spec.hs
iand675: Very good point. Will add tests for this.

utils/sqlcommenter/src/SqlCommenter.hs
jkachmar: I didn't see this sort of context in `newKey` docs, so it might be worth adding to a Haddock on `sqlCommenterKey` at least. up to you, though

utils/sqlcommenter/src/SqlCommenter.hs
jkachmar: aha okay, that's the missing link for me; I would maybe add this in a comment onto `unsafeConvert` in the same manner that Rust encourages documenting assumed invariants around `unsafe` blocks just so that it's clear.

michaelpj:
I think this could do with a bit of explanation for how to actually _use_ it? As far as I can tell, there are several things here:
1. Support for adding `sqlcommenter` comments to queries. This seems useful and arguably deserves its own library, which need not depend on any OTel bits?
2. Support for a named section of the thread-local context which is used for special attributes associated with `sqlcommenter`
3. Support for adding some OTel specific attributes into the special section

I guess the expectation is that on top of this you:
1. Modify the instrumentation libraries for databases to call `getSqlCommenterAttributesWithTraceData` and then use `sqlcommenter` to actually put them in queries.
2. Tell people to add things into the special `sqlcommenter` section in order to get them into their queries.

I guess I'm a bit unsure about the "action-at-a-distance" transmission of the attributes. This requires knowing to put attributes in a special place, that is a) specific to databases, and b) specific to a particular method of adding information into queries. As a user, I maybe don't even know that the `postgresql-simple` instrumentation library is using `sqlcommenter` - should I need to know that? If we were going to have a special section for such attributes, perhaps it should be called `postgres` or `database`?

An alternative would be to use something like the existing support for ["carry-on" attributes](https://hackage.haskell.org/package/hs-opentelemetry-api-0.2.0.0/docs/OpenTelemetry-Contrib-CarryOns.html). Then if a higher-scope piece of code wants to add an attribute that should get into the attributes of a database span, it can add it to the carry-ons?

michaelpj:
We discussed this at Zurihac, let me try and summarise what we concluded:
- The `sqlcommenter` behaviour looks a lot like a kind of `Propagator`
- At the moment, our ability to globally configure propagators is a bit limited: we can only configure ones that propagate into HTTP headers, as opposed to, say, queries.
- If we could globally configure "query propagators" then we could have a nice workflow:
    - Database instrumentation libraries could look up any appropriate configured query propagators and apply them
         - What does this look like? do we need per-database-type propagators? probably? Or could they just be key-value propagators that get then applied with `sqlcommenter`?
    - If you want to use `sqlcommenter`, you can add a query propagator that looks up the special attributes in the context and puts them in. So the spookiness is isolated to the definition of this propagator, which is optional.
    - If you want to do something else with `sqlcommenter`, you can write your own propagator that does something different.
- One way to configure the propagators in a more forwards-compatible way would be to have a `vault` of propagators, so different libraries can have well-known keys that let you look up specific propagators if they're present.

I am wondering whether it would indeed be nice to consider the query propagators as just key-value map propagators. Then `sqlcommenter` is just the implementation of how those maps then get put into the actual query. I think I need to take another look at what the spec says!

jkachmar:
superseded by #311

PR #191, opened 2025-10-02
chore: update nix.lock and run fourmolu

wenkokke:


wenkokke:
> The OTLP protobuf package is codegenned, so we intentionally don't format it with fourmolu. Could you drop those file modifications?

I've dropped those modifications and included a patch to the Makefile which excludes those files from the `make format` tasks.

iand675 (approved):

marinelli:
> > The OTLP protobuf package is codegenned, so we intentionally don't format it with fourmolu. Could you drop those file modifications?
> 
> I've dropped those modifications and included a patch to the Makefile which excludes those files from the `make format` tasks.

We should exclude `semantic-conventions/src`, the Haskell code inside it is generated [1] with [2].

[1] https://github.com/iand675/hs-opentelemetry/blob/3a429b9f09e88a6621bb83ffd9b47f9589948484/semantic-conventions/src/OpenTelemetry/SemanticConventions.hs#L3
[2] https://github.com/iand675/hs-opentelemetry/blob/3a429b9f09e88a6621bb83ffd9b47f9589948484/semantic-conventions/dev/generate.hs

PR #197, opened 2025-10-06
Add gRPC support to OTLP span exporter

wenkokke:
Follow up on #196 (includes all the same commits and 4e5a0fb7d83332d5877ae6782ba79c82b213fbc5) which adds gRPC support to `hs-opentelemetry-exporter-otlp`.

This currently depends on a fork of `grapesy`, since the `SpanExporter` type uses a `open/do/close` structure for resource management and `grapesy` only supports `withConnection`.


exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Span.hs
```diff
@@ -248,6 +270,176 @@ putWarningLn :: (MonadIO m) => String -> m ()
...
+      , Just uriAuthority <- uriAuthority uri
+      , Just addressPort <- readMaybe (uriPort uriAuthority) ->
+          pure $
+            G.Address
+              { G.addressHost = "localhost"
+              , G.addressPort = addressPort
+              , G.addressAuthority = Nothing
+              }
+      | otherwise -> do
+          putWarningLn $ "Warning: could not parse endpoint '" <> endpoint <> "'"
```
michaelpj: I don't think we should be just randomly polluting stderr like this. I see we do it already though.

michaelpj (approved):
Looks plausible! Have you tested it?

PR #198, opened 2025-10-07
Refactor SpanExporter API

wenkokke:
The current span exporter API doesn't quite match the protobuf API, which means that if you have data that is already grouped by resource and instrumentation library, you'll have to ungroup it in order to pass it to the exporter, which will then regroup it.

This PR defines `MaterializedResourceSpans`, `MaterializedScopeSpans`, and `MaterializedSpan`, and redefines the type of `SpanExporter` to accept `MaterializedResourceSpans` instead of `HashMap InstrumentationLibrary (Vector ImmutableSpan)`.


michaelpj (commented):
I'm mostly unsure about whether we need both `ImmutableSpan` and `MaterializedSpan`

PR #210, opened 2025-12-22
Make BoundedMap.itemMap strict

TeofilC:
I noticed that BoundedMap.itemMap was accumulating chains of thunks while looking at a heap with ghc-debug.

This isn't necessarily a problem. BoundedMap will eventually get forced. So, this cannot lead to a space leak.

However, this is unnecessarily laziness. There is a performance overhead to creating and evaluating thunks. Just making this strict should be a mild improvement for both allocations and runtime.

### DIFF
--- a/sdk/src/OpenTelemetry/Processor/Batch/Span.hs
+++ b/sdk/src/OpenTelemetry/Processor/Batch/Span.hs
@@ -166,7 +166,7 @@ data BoundedMap a = BoundedMap
   { itemBounds :: !Int
   , itemMaxExportBounds :: !Int
   , itemCount :: !Int
-  , itemMap :: HashMap InstrumentationLibrary (Builder.Builder a)
+  , itemMap :: !(HashMap InstrumentationLibrary (Builder.Builder a))
   }
 
 


michaelpj (approved):

michaelpj:
@iand675 looks like the required checks on PRs are out-of-date - I don't have permissions to alter the branch protection so I can't fix it.

iand675:
I'm going to close this branch, as the repo has had some heavy work done on it in the mean time. I'll mention that I benchmarked this idea across a variety of workloads, and have decided to not implement it. One principle I have for the OTel project is that it should be as low-overhead as possible when not actively used, and this forces all of the attribute formatting to run on the thread that's creating the traces. If we keep it lazy, then the exporter instead can force the thunks. In practice, this leads to significantly faster instrumentation callsites at the expense of a little bit of memory overhead.

TeofilC:
Sounds reasonable. This was mostly motivated by making it easier to reason about space leaks while looking at profiles, but I think you've fixed the space leak now in thread-utils

TeofilC:
> One principle I have for the OTel project is that it should be as low-overhead as possible when not actively used, and this forces all of the attribute formatting to run on the thread that's creating the traces. If we keep it lazy, then the exporter instead can force the thunks. In practice, this leads to significantly faster instrumentation callsites at the expense of a little bit of memory overhead.

FWIW in that case perhaps it would be better to use a different data structure here with cheaper appending, eg, a strict list of `InstrumentationLibrary` `Builder` pairs. That's what the heap representation will look like anyway. Explicitly making it that would likely make it easier to reason about this code (and perhaps by being explicit you can squeeze out some allocation savings). Then you can convert it to `HashMap InstrumentationLibrary Builder` in the exporter thread.

That would trade-off implementation complexity for not making it seem like there are space leaks here. And would also make it less likely that someone will accidentally break what sounds like a performance critical laziness invariant.

PR #136, opened 2024-07-09
Added Simple LogRecordProcessor and refactored shutdown and forceFlush

evanlauer1:

PR #211, opened 2026-01-03
Implement OTEL log exporter

ners:
### High-level overview
- Expose shared configuration functions outside of Trace-specific modules
- Add eventName to the log record
- Implement OTLP exporter for log records

Developed together with @vkleen and @mrcjkb. We've tested the log and trace exporters with the Grafana stack (Alloy, Tempo, Loki).

### Future work
- There is probably more opportunities for sharing between span and log exporters
- Respect the OTEL_EXPORTER_OTLP_*_ENDPOINT (currently ignored by trace exporter, too)
- More log exporters?

### AI disclosure
No AI/LLM was used in the development or preparation of this PR. Everything was copy-pasted by hand. :smile:


michaelpj:
I haven't looked into the logging stuff in much detail. Just wondering how this relates to https://github.com/iand675/hs-opentelemetry/pull/136, which looks like it got stuck in-flight?

api/src/OpenTelemetry/Internal/Logs/Types.hs
```diff
@@ -326,6 +305,9 @@ data ImmutableLogRecord = ImmutableLogRecord
   --
   -- In the contexts where severity participates in less-than / greater-than comparisons SeverityNumber field should be used.
   -- SeverityNumber can be compared to another SeverityNumber or to numbers in the 1..24 range (or to the corresponding short names).
+  , logRecordEventName :: Maybe Text
```
michaelpj: (This is a new field from the spec)

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Config.hs
```diff
@@ -0,0 +1,183 @@
+{-# LANGUAGE LambdaCase #-}
```
michaelpj: (This is a pure move, I think?)

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+--------------------------------------------------------------------------------
+
+{- |
+Internal helper.
+Construct a `LogRecordExporter` that uses HTTP/Protobuf.
+-}
+httpOtlpExporter :: (MonadIO m) => OTLPExporterConfig -> m LogRecordExporter
+httpOtlpExporter conf = do
+  -- TODO url parsing is jankym
+  -- TODO make retryDelay and maximum retry counts configurable
```
michaelpj: Copied todos from the span exporter. I guess we don't need to fix these now, but it does suggest there's some code that could be shared.

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+        req
+          { method = "POST"
+          , requestHeaders = encodingHeaders <> httpBaseHeaders conf req
+          , responseTimeout = httpTracesResponseTimeout conf
+          }
+  liftIO $
+    mkLogRecordExporter
+      LogRecordExporterArguments
+        { logRecordExporterArgumentsExport = \logs_ -> do
+            let anySpansToExport = not $ Vector.null logs_
```
michaelpj: s/spans/logs

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+                  pure $ Failure Nothing
+                else pure Success
+
+
+{- |
+Internal helper.
+Get the HTTP `ResponseTimeout` from the `OTLPExporterConfig`.
+-}
+httpTracesResponseTimeout :: OTLPExporterConfig -> ResponseTimeout
+httpTracesResponseTimeout conf = case otlpTracesTimeout conf <|> otlpTimeout conf of
```
michaelpj: surely we shouldn't be using the traces timeout, but rather some log timeout?

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+                  print resp
+                  pure $ Failure Nothing
+                else pure Success
+
+
+{- |
+Internal helper.
+Get the HTTP `ResponseTimeout` from the `OTLPExporterConfig`.
+-}
+httpTracesResponseTimeout :: OTLPExporterConfig -> ResponseTimeout
```
michaelpj: wrong name. Please do a grep for rogue occurences of "trace" in the log files!

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+  where
+    responseTimeoutMilli :: Int -> ResponseTimeout
+    responseTimeoutMilli = responseTimeoutMicro . (* 1_000)
+
+
+{- |
+Internal helper.
+Get the HTTP host from the `OTLPExporterConfig`.
+-}
+httpHost :: OTLPExporterConfig -> String
```
michaelpj: I think maybe some of these should go in OTLP.Config?

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -1,2 +1,365 @@
...
+  concat
+    [ [(hContentType, httpProtobufMimeType)]
+    , [(hAcceptEncoding, httpProtobufMimeType)]
+    , fromMaybe [] (otlpHeaders conf)
+    , fromMaybe [] (otlpTracesHeaders conf)
+    , requestHeaders req
+    ]
+
+
+--------------------------------------------------------------------------------
```
michaelpj: There's a lot of duplicated code here. It would be much easier to review this if that was factored out so we can focus on the new stuff.

michaelpj (commented):
This PR does a number of things and it's not clear which ones are essential or not. Would it be possible to split it up into a number of stages?

In particular, there are a number of moves and minor refactors, which make it harder to review and need to be assessed a bit on their own terms (e.g. are the moves API breakages?).

I also think it would be worth the effort to try and reduce the code duplication. Possibly that looks like abstracting over traces/logs a bit. Fundamentally both trace and log processing involves handling items which can be turned into protobuf and then sending them to some endpoint.

Finally, it would be nice to have _some_ tests. This project is not terribly well tested, but there are _some_ tests which check that e.g. we can run with a trace processor that accumulates traces in memory, and then when we look at it we get the traces we expect. It would be good to have corresponding tests for the logging functionality.

ners:
@michaelpj @iand675 any chance we can make progress with this PR?

michaelpj:
It looks like @iand675 has landed a version of log support?

PR #217, opened 2026-04-10
Full OTel 1.55.0 compliance, performance improvements

iand675:
This branch is the accumulated work toward a (formerly 0.4, but actually merits being) 1.0 release of the Haskell OpenTelemetry packages. It touches essentially every package in the repo. Many of the API changes are _deeply_ intertwined, and I am loath to do frequent compatibility churn on consumers of the project, so I feel like in this case it is best to fully modernize in terms of spec conformance and performance in one shot.

### The spec was way behind

I did a systematic audit (pretty much going over every MUST and SHOULD and MUST NOT and SHOULD not statement in all of the relevant OTel specs) against OTel spec 1.55.0, OTLP 1.10.0, and semconv 1.40.0. The previous codebase predated spec stabilization, so there were quite a few things that were super outdated (up to 5 years out of date!), such as:

- `InstrumentationScope` (spec renamed from InstrumentationLibrary ages ago)
- `TraceState.lookup` not being implemented
- `Span.isValid` requiring both IDs non-zero
- `SpanExporter.forceFlush` behavior
- Logs `EventName` + `Logger.Enabled`
- Proper shutdown / forceFlush propagation across all three signals were all missing or wrong. 

The sampler interface now receives the `InstrumentationScope`, which is a breaking change for anyone with a custom `Sampler`, but the spec says it is a `MUST`, so here we are.

### Performance was leaving a lot on the table

The old span hot path was doing a depressing amount of work per span:

- Every `getSpanContext` went through an `IORef` indirection even though trace/span IDs never change after creation
- `TraceId`/`SpanId` were heap-allocated using `ShortByteString`s which are pinned & GC-visible instead of using register-width unpacked `Word64`s.
- ID generation used `System.Random.Stateful` which has global lock contention issues.
- `getContext` allocated a `ThreadId` box via FFI, did an O(log n) `IntMap.lookup` inside a 32-stripe array, then `lookupSpan` went through a `Data.Vault.Strict` `HashMap` lookup
- Timestamps went through the `clock` package's `alloca`/`errno`/`Storable` logic.

I did quite a lot of benchmarking in order to get performance into a good state, and I'm quite pleased. After all of the changes, we beat Rust and Go in terms of instrumentation overhead for many use cases. The new implementation does the following, among other things:

- Splits `Span` into immutable identity fields (accessed directly instead of via IORef) and a `SpanHot` sub-record behind one IORef for mutable state
- Uses unboxed `Word64` pairs for TraceId/SpanId
- Uses a thread-local xoshiro256++ in C for ID generation (seeded once from platform CSPRNG, then zero syscalls / zero Haskell allocation)
- Introduces constructors for the most common code paths that allow for case-of-case to give significantly better codegen results than always going through record-of-function invocations (e.g. for codegen, sampling, etc)
- Custom CMM primops that fuse `CurrentTSO.id` reads with a flat hash table probe. This brings `getContext` from 17.3ns to 2.9ns, and `lookupSpan` from 10.0ns to 0.6ns
- Direct `clock_gettime` FFI for timestamps (Word64 nanoseconds), which allows us to avoid allocating pinned byte arrays for just a single `Word64` that we care about
- C SIMD/SWAR hex encoding for trace/span IDs
- `inSpan` skips mask/bracket/context entirely when no processors are registered
- The thread-local context rewrite lives in a local `thread-utils-context` repo (the `StorageV2` module hasn't been published to Hackage yet, which is why the stack files reference `../thread-utils/`). Obviously need to fix that before shipping, but we can start code review.

### Other notable changes

Metrics didn't exist, and logs barely existed

The metrics API and SDK are entirely new: views, exemplars, cardinality guards, advisory attributes, Prometheus + OTLP + in-memory exporters. 

The logs SDK is also refreshed to better conform to the spec as well as provide some ecosystem integration with e.g. katip and co-log. Both logs and metrics follow the same patterns as the trace SDK.

OTLP protos were updated to v1.10.0 with regenerated Haskell bindings. Semantic conventions regenerated to 1.40.0.

### New packages

- **Propagators**: Datadog, X-Ray, Jaeger (these existed in various states of incompleteness before; now they're tested and documented)
- **Instrumentation**: `ghc-metrics`, `katip`, `co-log`, `monad-logger`, Cloudflare (with Turnstile support)
- **`hs-opentelemetry-api-types`**: leaf package with just `Attribute`/`AttributeKey` types so downstream libs can depend on attributes without pulling in the full API
- **Declarative configuration**: `OTEL_CONFIG_FILE` support (YAML)
- **Resource detectors**: AWS EC2/ECS/EKS, GCP Compute, Azure VM, Heroku, container, FaaS are all auto-detected at SDK init

### Build fixes
The `connection` package (not the crypton fork) was still listed as a direct dep in the SDK. It's not in Stackage LTS-24+ and can't build against `tls 2.x` in nixpkgs. Swapped to `crypton-connection` / `crypton-x509-store` which are drop-in replacements. Removed the `doJailbreak` hack from the nix overlay. Stack and nix both resolve cleanly now.

Also went through and fixed every GHC warning across the repo.

Ensured GHC 9.12 support.

### Documentation

All ~30 subpackage READMEs were rewritten to be more didactic: one-paragraph description, usage snippet, link back to root. The root README has a progressive intro with before/after code examples, getting started sections for each signal type. There's a full migration guide in `docs/migration-guide.md` covering every breaking change that I could remember after all of these changes.

## Breaking changes (the short version)

The full list is in `docs/migration-guide.md`. The highlights:

- `Span` internals restructured (direct `ImmutableSpan` instead of `IORef ImmutableSpan`)
- `TraceId`/`SpanId` are `Word64` (not `ShortByteString`)
- `IdGenerator` and `Sampler` are ADTs (not records); sampler takes scope param
- `Timestamp` is `Word64` nanoseconds (not `TimeSpec`)
- `Context` has dedicated slots for `Span` and `Baggage` (not `Vault`)
- `Propagator` uses `TextMap` carrier (not `RequestHeaders`)
- Processor and exporter result types changed
- Module renames: `Logs` → `Log`, `Metrics` → `Metric`
- Resource schema phantom type removed

Most of these are internal types that end users don't touch directly, so if you're just using `inSpan` and `withTracerProvider`, the migration is mostly import path changes.

## Test plan

- [ ] Build is clean with 0 GHC warnings across all supported ghc versions
- [ ] Nix environment works
- [ ] All examples work
- [ ] Full test suite pass across all versions
- [ ] Try running against a large production system, make sure nothing funky crops up


api/src/OpenTelemetry/Internal/Common/Types.hs
```diff
@@ -151,45 +225,128 @@ instance ToValue AnyValue where
...
-  -- First try and parse with a mandatory version string on the end. If that fails, try
-  -- to parse just a package name
-  let fullRegex :: String = "(" <> packageNameRegex <> "-" <> versionRegex <> ")|" <> packageNameRegex
-  (_ :: String, _ :: String, _ :: String, groups :: [String]) <- packageString =~~ fullRegex
-  -- We end up with 5 groups overall
-  (name, version) <- case groups of
-    [_, name, version, ""] -> pure (name, version)
-    [_, _, _, name] -> pure (name, "")
-    _ -> fail $ "could not parse package string: " <> packageString
-  pure $ InstrumentationLibrary {libraryName = T.pack name, libraryVersion = T.pack version, librarySchemaUrl = "", libraryAttributes = emptyAttributes}
```
iand675, line 189: Specifically here I'm replacing this so that we can drop the `regex-tdfa` dependency for a relatively minor use-case

PR #219, opened 2026-04-15
Metrics implementation (API + SDK + exporters + build scaffolding)

iand675:
#### 🥞 PR Stack (1/18)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#236](https://github.com/iand675/hs-opentelemetry/pull/236) Comprehensive test suite, examples, vendors, build infra finalization
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#235](https://github.com/iand675/hs-opentelemetry/pull/235) Update existing instrumentation + add ghc-metrics, gogol, amazonka, logging bridges
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#234](https://github.com/iand675/hs-opentelemetry/pull/234) SDK updates — configuration, resource detectors, processors, traces
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#233](https://github.com/iand675/hs-opentelemetry/pull/233) Update handle, in-memory, OTLP, prometheus exporters for API 0.4
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#232](https://github.com/iand675/hs-opentelemetry/pull/232) Update w3c/b3/datadog, add jaeger and xray propagators
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#231](https://github.com/iand675/hs-opentelemetry/pull/231) Comprehensive test suite for API 0.4
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#230](https://github.com/iand675/hs-opentelemetry/pull/230) API 0.4 — restructured modules, FFI perf, api-types, semantic-conventions 1.40
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#229](https://github.com/iand675/hs-opentelemetry/pull/229) Update README, licenses, changelogs, migration guide, spec compliance
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#228](https://github.com/iand675/hs-opentelemetry/pull/228) Propagate shutdown/forceflush across all signals, fix barrel module exports
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#227](https://github.com/iand675/hs-opentelemetry/pull/227) Metrics follow-up fixes from log processor integration
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#226](https://github.com/iand675/hs-opentelemetry/pull/226) OTLP, handle, and in-memory log exporters
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#225](https://github.com/iand675/hs-opentelemetry/pull/225) Batch and simple LogRecordProcessor
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#222](https://github.com/iand675/hs-opentelemetry/pull/222) Add EventName field and Logger.Enabled API
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#223](https://github.com/iand675/hs-opentelemetry/pull/223) Add ForceFlush to SpanExporter (MUST per spec)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#224](https://github.com/iand675/hs-opentelemetry/pull/224) Add TraceState.lookup (MUST per spec)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#221](https://github.com/iand675/hs-opentelemetry/pull/221) Require both TraceId AND SpanId non-zero for isValid
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[#220](https://github.com/iand675/hs-opentelemetry/pull/220) Update OTLP protos to v1.10.0 and regenerate Haskell bindings
👉 **#219 Metrics implementation (API + SDK + exporters + build scaffolding)**

---

## Summary
- Add metrics API: Counter, Histogram, UpDownCounter, Gauge instruments
- Add MeterProvider + Meter to SDK with exponential histogram aggregation
- Add Prometheus exporter for metrics
- Update stack build configuration for metrics packages

Part 1 of 18 in the hs-opentelemetry v0.4 stack.


exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Metric.hs
```diff
@@ -0,0 +1,487 @@
...
+
+
+type Encoder = L.ByteString -> L.ByteString
+
+
+httpMetricsCompression :: OTLPExporterConfig -> ([(HeaderName, C.ByteString)], Encoder)
+httpMetricsCompression conf =
+  case otlpMetricsCompression conf <|> otlpCompression conf of
+    Just GZip -> ([(hContentEncoding, "gzip")], compress)
+    _ -> ([], id)
```
jkachmar, line 176: any objection to adding `zstd` in a future PR or is there a reason this has to be `gzip`?

exporters/otlp/test/Spec.hs
```diff
@@ -1,3 +1,309 @@
...
+          V.length v `shouldBe` 2
+          metricNameAt v 0 `shouldBe` Just "metric-a"
+          metricNameAt v 1 `shouldBe` Just "metric-b"
+
+
+firstMetric :: ExportMetricsServiceRequest -> Maybe PM.Metric
+firstMetric r = do
+  rm <- (r ^. MSF.vec'resourceMetrics) V.!? 0
+  sm <- (rm ^. Mf.vec'scopeMetrics) V.!? 0
+  (sm ^. Mf.vec'metrics) V.!? 0
```
jkachmar, line 256: nit (not worth disrupting the PR chain, maybe worth following up in the future): these feel like they'd benefit from being written as optics directly

exporters/prometheus/src/OpenTelemetry/Exporter/Prometheus.hs
```diff
@@ -0,0 +1,390 @@
...
+renderScope :: Map.Map Text Text -> ScopeMetricsExport -> Text
+renderScope resMap ScopeMetricsExport {..} =
+  let jobMap =
+        if T.null (libraryName scopeMetricsScope)
+          then resMap
+          else Map.insert "job" (libraryName scopeMetricsScope) resMap
+  in T.intercalate "\n" $ fmap (renderMetric jobMap) $ V.toList scopeMetricsExports
+
+
+renderMetric :: Map.Map Text Text -> MetricExport -> Text
```
jkachmar, line 67: would it be much of a contortion to use a `Builder` for these, or do you think that wouldn't be worthwhile?

sdk/src/OpenTelemetry/MeterProvider.hs
```diff
@@ -0,0 +1,1252 @@
...
+  st <- readIORef (sdkMeterStorage env)
+  t <- nowNanos
+  let temp = sdkMeterAggregationTemporality env
+      snap = storageCells st
+      startT = sdkMeterStartTimeNanos env
+      rme = buildResourceExport (sdkMeterResource env) startT t temp snap
+  _ <-
+    if temp == AggregationDelta
+      then atomicModifyIORef' (sdkMeterStorage env) $ \s -> (applyDeltaReset s, ())
+      else pure ()
```
jkachmar, line 1054:
```suggestion
  when (temp == AggregationDelta) do
    atomicModifyIORef' (sdkMeterStorage env) $ \s -> (applyDeltaReset s, ())
```

sdk/src/OpenTelemetry/MeterProvider.hs
```diff
@@ -0,0 +1,1252 @@
...
+  pure [rme]
```
jkachmar, line 1055: why does this return a list if it's only ever going to return a single item?

sdk/src/OpenTelemetry/MeterProvider.hs
```diff
@@ -0,0 +1,1252 @@
...
+                        , sumDataPointTimeUnixNano = t
+                        , sumDataPointValue = if scIsInt sc then Left (round (scValue sc)) else Right (scValue sc)
+                        , sumDataPointAttributes = applyDimAttrs d attrs
+                        , sumDataPointExemplars = scExemplars sc
+                        }
+                        : acc
+                    _ -> acc
+                )
+                []
+                series
```
jkachmar, line 1129: this pattern's repeated a bunch here; feels like it could be extracted into a helper that gets reused among these export functions

semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
```diff
@@ -4,6 +4,7 @@ name: hs-opentelemetry-semantic-conventions
 version: 0.1.0.0
 author: Kazuki Okamoto (岡本和樹)
 maintainer: kazuki.okamoto@herp.co.jp
+license: BSD-3-Clause
```
jkachmar, line 7: if this was contributed by someone else changing the license is going to need affirmation

semantic-conventions/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright Ian Duncan (c) 2021
```
jkachmar, line 1: copyright should be updated & attribution should reflect the package author

api/src/OpenTelemetry/Metrics.hs
```diff
@@ -0,0 +1,168 @@
...
+  Gauge (..),
+  ObservableResult (..),
+  ObservableCallbackHandle (..),
+  ObservableCounter (..),
+  ObservableUpDownCounter (..),
+  ObservableGauge (..),
+
+  -- * Instrument name validation (for SDK implementers)
+  InstrumentName.validateInstrumentName,
+  InstrumentName.validateInstrumentUnit,
```
jkachmar, line 39: what's the tradeoff on validating this vs. `newtype`-wrapping w/ a parse function? is there a situation where we would want to let people construct instruments w/out validation?

api/src/OpenTelemetry/Metrics.hs
```diff
@@ -0,0 +1,168 @@
...
+  MeterProvider
+    { meterProviderGetMeter = \scope -> pure (noopMeter scope)
+    , meterProviderShutdown = pure ShutdownSuccess
+    , meterProviderForceFlush = pure FlushSuccess
+    }
+
+
+globalMeterProvider :: IORef MeterProvider
+{-# NOINLINE globalMeterProvider #-}
+globalMeterProvider = unsafePerformIO $ newIORef noopMeterProvider
```
jkachmar, line 160:
```suggestion
globalMeterProvider :: IORef MeterProvider
globalMeterProvider = unsafePerformIO $ newIORef noopMeterProvider
{-# NOINLINE globalMeterProvider #-}
```

jkachmar:
high-level comments:
1. a lot of these modules have both `StrictData` and strictness annotations on the data types; we should pick one and stick with it
2. there's a lot of verbosity in record names & repetition in transforming between stuff; some of that's going to be unavoidable, but i'm a little wary of how it plays into the public API

**(1)** is easy to clean up and **(2)** is the sort of thing we can discuss after the stack has been merged & we can read it all in the context of examples & whatnot

jkachmar (approved):

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Metric.hs
iand675: No objection in particular

semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
iand675: @kakkun61 It looks like this package currently has no license set. Would it be alright if I set the SPDX license identifier (BSD-3-Clause) in the cabal file, etc.?

Having the license explicitly specified is very helpful for users.

Thank you.

exporters/otlp/test/Spec.hs
iand675: We're using lens-family, and it doesn't really have a great option to tidy this up further that I can tell. 


```
-- | Traversal into the first element of a Vector (if non-empty).
_head :: Traversal' (V.Vector a) a
_head f v = case v V.!? 0 of
  Nothing -> pure v
  Just a -> (\a' -> v V.// [(0, a')]) <$> f a
```

I'd have to do something like that, which feels gross for just a test.

api/src/OpenTelemetry/Metrics.hs
iand675: This is a spec conformance thing:

  ▎ The API MUST treat instrument name as opaque. The SDK SHOULD validate instrument names against the instrument name syntax.

So the current design is correct in terms of spec conformance. We do `Text` within API, validation as a SDK-side concern. The API is supposed to accept anything; if the SDK decides the name is invalid, it returns a noop instrument (which is exactly what happens in validateOrNoop).

iand675:
Re:

> there's a lot of verbosity in record names & repetition in transforming between stuff; some of that's going to be unavoidable, but i'm a little wary of how it plays into the public API

The rest of the library already skews pretty heavily towards just being verbose, in part because of the spec, and in part because the library itself predates me feeling comfortable with duplicate record fields and so forth.

I'd probably prefer to just be verbose for consistency's sake at this point to match the rest of the API/SDK style

sdk/src/OpenTelemetry/MeterProvider.hs
iand675: `collectResourceMetrics` returns [ResourceMetricsExport] because that's what `metricExporterExport` expects (matching the OTLP repeated ResourceMetrics shape). Right now it's always a singleton since we have one resource per provider, but the spec's `MetricProducer` concept (external metric sources that contribute additional ResourceMetricsExport entries with potentially different resources) would make use of the list. We haven't wired that up yet, but the type is forward-compatible with it.

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/Metric.hs
iand675: (just to be clear, I'll probably want to circle back and implement this later on instead of now)

jkachmar (approved):

semantic-conventions/hs-opentelemetry-semantic-conventions.cabal
kakkun61: It's OK! I'm sorry for my very late reply.

PR #252, opened 2026-05-14
feat(api): API 0.4 - restructured modules, FFI perf, api-types, semantic-conventions 1.40, build scripts

iand675:
Part of the hs-opentelemetry v0.4 stack. Stacked on #251.

See PR #219 for the base of this stack.

Made with [Cursor](https://cursor.com)


api/src/OpenTelemetry/Context/ThreadLocal.hs
```diff
@@ -55,14 +90,64 @@ module OpenTelemetry.Context.ThreadLocal (
...
+emptyEntry = ContextEntry empty 0
+{-# INLINE emptyEntry #-}
+
+
+{- | Opaque token returned by 'attachContext'.
+
+Pass this to 'detachContext' to restore the context that was active before
+the attach. The implementation validates LIFO ordering: detaching out of
+order (i.e. with a token that doesn't match the most recent attach) logs
+a diagnostic error per the OpenTelemetry specification.
```
jkachmar, line 124: this is awesome; i didn't realize this was per the spec

api/src/OpenTelemetry/Internal/Log/Types.hs
```diff
@@ -0,0 +1,581 @@
...
+  -- ^ Time when the event occurred. 'UNothing' when unknown.
+  , logRecordObservedTimestamp :: !Timestamp
+  {- ^ Time when the event was observed by the collection system. For events that originate in OpenTelemetry (e.g. using OpenTelemetry Logging SDK)
+  this timestamp is typically set at the generation time and is equal to Timestamp. For events originating externally and collected by OpenTelemetry (e.g. using Collector)
+  this is the time when OpenTelemetry’s code observed the event measured by the clock of the OpenTelemetry code. This field SHOULD be set once the event is observed by OpenTelemetry.
+
+  For converting OpenTelemetry log data to formats that support only one timestamp or when receiving OpenTelemetry log data by recipients that support only one timestamp internally the following logic is recommended:
+  - Use Timestamp if it is present, otherwise use ObservedTimestamp
+  -}
+  , logRecordTracingDetails :: {-# UNPACK #-} !(UMaybe (TraceId, SpanId, TraceFlags))
```
jkachmar: maybe a dumb question but if you're using `UMaybe` here would it be worth unpacking the tuple as well or does GHC automatically optimize that if the entries are small enough?

api/src/OpenTelemetry/Internal/Log/Types.hs
```diff
@@ -0,0 +1,581 @@
...
+  | Warn4
+  | Error
+  | Error2
+  | Error3
+  | Error4
+  | Fatal
+  | Fatal2
+  | Fatal3
+  | Fatal4
+  | Unknown !Int
```
jkachmar, line 498: does the spec mandate that we support this? i feel like the only special case we might want to support would be:
> SeverityNumber=0 MAY be used to represent an unspecified value.

...but if we _do_ restrict ourselves to the valid severity values then we could derive `Bounded` as well which feels kind of useful.

idk, i suppose if we permit an `Unknown` field it makes conversions to/from `Int` easier in that we don't have to support an optional failure?

api/src/OpenTelemetry/Internal/Metric/Export.hs
```diff
@@ -25,41 +42,85 @@ import Data.Text (Text)
...
+
+{- | An optional 'Double', using UNPACK to avoid the extra box that @Maybe Double@ incurs.
+2 words for 'SomeDouble' vs 4 for @Just (D# x)@.
+
+@since 0.0.1.0
+-}
+data OptionalDouble
+  = NoDouble
+  | SomeDouble {-# UNPACK #-} !Double
+  deriving stock (Eq, Show, Generic)
```
jkachmar, line 79: why this vs. `UMaybe Double`?

api/src/OpenTelemetry/Internal/Trace/Id.hs
```diff
@@ -25,60 +28,82 @@ module OpenTelemetry.Internal.Trace.Id (
...
+import System.IO.Unsafe (unsafeDupablePerformIO)
+import Prelude hiding (length)
 
 
-#if MIN_VERSION_base(4,17,0)
-import GHC.Exts (word64ToWord#)
-#endif
+-- ---------------------------------------------------------------------------
+-- C FFI
+-- ---------------------------------------------------------------------------
```
jkachmar: small nit (not necessary to change in this PR if you think it'd mess up the stack): instead of using these `-----` blocks might it be nice to have Haddock section headings?

api/src/OpenTelemetry/Internal/Trace/Id.hs
```diff
@@ -150,126 +155,294 @@ traceIdBytes (TraceId bytes) = fromShort bytes
...
+    :: Word64
+    -> Word64
+    -> Word64
+    -> Word8
+    -> Word8
+    -> Ptr Word8
+    -> IO ()
+
+
+{- | Parse a W3C traceparent header in a single SIMD-accelerated C call.
```
jkachmar, line 404: sick

api/src/OpenTelemetry/Resource/Process.hs
```diff
@@ -1,10 +1,4 @@
 {-# LANGUAGE CPP #-}
```
jkachmar: do we need `CPP`?

api/src/OpenTelemetry/Baggage.hs
```diff
@@ -116,21 +178,164 @@ data Property = Property
...
+isTokenChar :: Char -> Bool
+isTokenChar c = w < 128 && tokenTable `BS.index` fromIntegral w /= 0
+  where
+    w = fromEnum c
+{-# INLINE isTokenChar #-}
+
+
+-- 128-byte lookup: 1 = valid token char, 0 = invalid
+tokenTable :: ByteString
+tokenTable =
```
jkachmar, line 207: i guess this is never exported so it doesn't really need to be a haddock; shame there isn't a better way to lay out the LUT

api/src/OpenTelemetry/Baggage.hs
```diff
@@ -179,93 +387,177 @@ token =
...
+                  (valBs, rest3) = BS.span isValueByte rest2
+              rest <- go (stripOWS rest3)
+              propVal <- case decodeUtf8' (percentDecode valBs) of
+                Right t -> Right t
+                Left _ -> Left "Invalid UTF-8 in baggage property value"
+              pure $ Property (Token keyBs) (Just propVal) : rest
+            else do
+              rest <- go rest1
+              pure $ Property (Token keyBs) Nothing : rest
+      | otherwise = Left $ "Unexpected byte in baggage: " ++ show (BS.head s)
```
jkachmar, line 518: what's the benefit of handling it this way vs. a parser combinator approach; lighter dependency footprint?

api/src/OpenTelemetry/Common.hs
```diff
@@ -1,13 +1,97 @@
...
+@Word64#@ directly in its closure: 2 words total vs 4 for
+@Just (Timestamp ns)@.
+
+@since 0.0.1.0
+-}
+data OptionalTimestamp
+  = NoTimestamp
+  | SomeTimestamp {-# UNPACK #-} !Word64
+  deriving (Eq, Ord)
+
```
jkachmar, line 45: same q as w/ `OptionalDouble`: is there a perf gain here vs. `UMaybe`?

api/src/OpenTelemetry/Environment.hs
```diff
@@ -69,6 +85,51 @@ lookupMetricExportIntervalMillis = do
...
+See <https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection>.
+
+@since 0.4.0.0
+-}
+data LogsExporterSelection
+  = LogsExporterNone
+  | LogsExporterOtlp
+  | LogsExporterConsole
+  | -- | A name not in the built-in set; looked up via the exporter registry.
+    LogsExporterCustom !String
```
jkachmar, line 109: `String` for convenience when pulling it from the env?

api/src/OpenTelemetry/Registry.hs
```diff
@@ -0,0 +1,350 @@
...
+{- | Register a span exporter factory, replacing any existing entry
+with the same name.
+
+Use this from third-party exporter packages to override or extend the
+set of exporters available to the SDK.
+
+@since 0.4.0.0
+-}
+registerSpanExporterFactory :: Text -> IO SpanExporter -> IO ()
+registerSpanExporterFactory = insertRegistry spanExporterRegistry
```
jkachmar, line 146: is the factory idiom part of the spec or is it mostly just the idiom that makes the most sense given what's being registered here?

api/src/OpenTelemetry/Util.hs
```diff
@@ -82,22 +91,132 @@ instance (Constructor c) => HasConstructor (C1 c f) where
...
+Performs a strict read, applies @f@ to decide both the new value and a
+pre-swap result, then CAS's the new value in. Returns the old value (before
+modification) on success. Retries on CAS failure.
+
+Used by 'endSpan' where we need to atomically set @spanEnd@ and also read the
+(unchanged) Tracer field to obtain the processor vector, all without
+navigating the Tracer inside an @atomicModifyIORef'@ closure.
+
+@since 0.0.1.0
+-}
```
jkachmar, line 180: feels like it might be useful to pull some of the atomics out into a separate package in the future

scripts/fix-stack-compilation.sh
```diff
@@ -0,0 +1,164 @@
+#!/bin/bash
```
jkachmar: might want to drop this, but up to you

jkachmar (approved):

api/src/OpenTelemetry/Common.hs
iand675: I'm reluctant to make UMaybe be part of the user-facing interface. My goal is for that to mostly be an internal detail, but not make people deal with a bunch of unlifted bullshit in the usual face of things.

api/src/OpenTelemetry/Baggage.hs
iand675: Yes, precisely. IIRC this let me drop an attoparsec dependency.

api/src/OpenTelemetry/Baggage.hs
iand675: Yeah this was a situation where I could make it a C import, maybe. but that felt like more work than I really wanted to do. The tradeoff is that fourmolu hates me forevermore.

api/src/OpenTelemetry/Environment.hs
iand675: Yeah, in practice this is only used to do custom dispatch on initializing some logger and using that, so I didn't really see the point in forcing it through text for a situation where it was used literally once at initialization time.

api/src/OpenTelemetry/Internal/Log/Types.hs
iand675: My goal here was that if someone ended up stuck on this version of the library but started having to deal with newer versions of the spec, they may end up needing to deal with custom numeric values that this version of code has no semantic understanding about. This is just a little escape hatch to accomodate forward-compatibility.

api/src/OpenTelemetry/Internal/Log/Types.hs
iand675: Good question! I'm not sure. Will poke at it a bit.

api/src/OpenTelemetry/Registry.hs
iand675: it's a spec thing yep

PR #299, opened 2026-05-28
Process resource: creation time, Opt-In fields, fix command_args

iand675:
## Summary
- process.creation.time (Required per spec): detected at SDK init
- process.command_args includes argv[0] per spec
- 6 new Opt-In fields with auto-detection: args_count, parent_pid, working_directory, interactive, title, linux.cgroup

## Test plan
- [x] cabal test hs-opentelemetry-sdk

### DIFF
--- a/sdk/src/platform/windows/OpenTelemetry/Platform.hs
+++ b/sdk/src/platform/windows/OpenTelemetry/Platform.hs
@@ -10,3 +10,7 @@ import qualified Data.Text as T
 
 tryGetUser :: IO (Maybe T.Text)
 tryGetUser = pure Nothing
+
+
+tryGetParentProcessID :: IO (Maybe Int)
+tryGetParentProcessID = pure Nothing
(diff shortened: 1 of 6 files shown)


sdk/src/platform/windows/OpenTelemetry/Platform.hs
```diff
@@ -10,3 +10,7 @@ import qualified Data.Text as T
 
 tryGetUser :: IO (Maybe T.Text)
 tryGetUser = pure Nothing
+
+
+tryGetParentProcessID :: IO (Maybe Int)
+tryGetParentProcessID = pure Nothing
```
jkachmar, line 16: windows supports this doesn't it? or do we just want to skip for now

jkachmar (approved):

sdk/src/platform/windows/OpenTelemetry/Platform.hs
iand675: I just don't have a windows machine set up to test on. Happy for someone to add support if they pinky promise that it works

PR #307, opened 2026-06-09
Attach backtraces from exception context with base-4.20+

### LINKED ISSUE #239: Exception call stacks don't include all possible call stack mechanisms
In fact, at the moment they use `whoCreated` which is only cost-center call stacks. So e.g. even `HasCallStack` call stacks will be lost.

michaelpj:
This uses the new backtraces mechanism to report whatever backtraces are
collected by GHC.

To make this work, we have to be a bit careful about `SomeException`.
Naively unwrapping it can discard the exception context, which we don't
want.

Fixes #239.

### DIFF
--- a/api/src/OpenTelemetry/Trace/Core.hs
+++ b/api/src/OpenTelemetry/Trace/Core.hs
@@ -210,6 +210,7 @@ module OpenTelemetry.Trace.Core (
 
   -- ** Recording error information
   recordException,
+  recordSomeException,
   recordError,
   setStatus,
   SpanStatus (..),
@@ -261,6 +262,17 @@ import Control.Concurrent.Async
 import Control.Concurrent.Thread.Storage (getCurrentThreadId)
 import Control.Exception (Exception (..), SomeException (..), catch, displayException)
 import qualified Control.Exception as EUnsafe
+
+
+#if MIN_VERSION_base(4,20,0)
+import Control.Exception (someExceptionContext)
+import Control.Exception.Annotation (displayExceptionAnnotation)
+import Control.Exception.Backtrace (Backtraces)
+import Control.Exception.Context (ExceptionContext, getExceptionAnnotations)
+#endif
+#if MIN_VERSION_base(4,21,0)
+import Control.Exception (WhileHandling)
+#endif
 import Control.Monad
 import Control.Monad.IO.Class
 import Control.Monad.IO.Unlift
@@ -694,9 +706,9 @@ inSpanInternal t n args extraAttrs f
           case classification of
             ErrorException -> do
               setStatus s $ Error $ T.pack $ displayException inner
-              recordException s (H.union [(unkey SC.exception_escaped, toAttribute True)] exAttrs) Nothing inner
+              recordSomeException s (H.union [(unkey SC.exception_escaped, toAttribute True)] exAttrs) Nothing someEx
             RecordedException ->
-              recordException s (H.union [(unkey SC.exception_escaped, toAttribute True)] exAttrs) Nothing inner
+              recordSomeException s (H.union [(unkey SC.exception_escaped, toAttribute True)] exAttrs) Nothing someEx
             IgnoredException ->
               pure ()
           endSpan s Nothing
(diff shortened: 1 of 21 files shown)


iand675:
This looks great! Paging @parsonsmatt for a second set of eyes on this, but I don't immediately see anything I'm concerned about.

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -1023,30 +1035,83 @@ endSpan (Dropped _) _ = pure ()
...
- @since 0.0.1.0
+Prefer using @recordSomeException@ to preserve exception context.
+
+@since 0.0.1.0
 -}
 recordException :: (MonadIO m, Exception e) => Span -> AttributeMap -> Maybe Timestamp -> e -> m ()
-recordException s attrs ts e = liftIO $ do
-  cs <- whoCreated e
-  let message = T.pack $ displayException e
+recordException s attrs ts e = recordSomeException s attrs ts (toException e)
```
parsonsmatt, line 1044: i really cannot get over how much i dislike this design decision from the core GHC. if we can possibly handle this *not* swallowing all annotations, then that would be a huge win. As it stands, every use of `recordException` is going to continue throwing away annotations.

is there really not a `toExceptionWithContext` we can use? or some other design pattern? Perhaps we can require `(HasExceptionContext) => ` as a constraint, and if we're in a branch that unpacks `SomeException`, that should auto solve. ie `case someException of SomeException exn -> toExceptionWithContext exn` ought to work?

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -1023,30 +1035,83 @@ endSpan (Dropped _) _ = pure ()
...
     NewEvent
       { newEventName = "exception"
       , newEventAttributes =
           H.union
             attrs
-            [ (unkey SC.exception_type, A.toAttribute $ T.pack $ show $ typeOf e)
+            [ (unkey SC.exception_type, A.toAttribute $ T.pack $ show $ typeOf inner)
             , (unkey SC.exception_message, A.toAttribute message)
-            , (unkey SC.exception_stacktrace, A.toAttribute $ T.unlines $ map T.pack cs)
+            , (unkey SC.exception_stacktrace, A.toAttribute stack)
```
parsonsmatt, line 1077: nice!

should we try to do anything to capture any other annotations on the SomeException?

api/src/OpenTelemetry/Trace/Core.hs
```diff
@@ -1023,30 +1035,83 @@ endSpan (Dropped _) _ = pure ()
...
+exceptionStackText se = pure $ T.pack $ unlines $ filter (not . null) sections
+  where
+    annotations :: ExceptionContext
+    annotations = someExceptionContext se
+
+    sections :: [String]
+    sections =
+      -- We use 'displayExceptionAnnotation' so we get the same rendering as GHC
+      -- would use
+      map displayExceptionAnnotation (getExceptionAnnotations annotations :: [Backtraces])
```
parsonsmatt: a type application may be slightly more legible here if it's in style guide for the project

api/test/OpenTelemetry/Trace/ExceptionHandlerSpec.hs
```diff
@@ -162,6 +168,135 @@ spec = describe "ExceptionHandler" $ do
...
+-- 'Backtraces' annotation populated from 'HasCallStack' on GHC 9.10+, which
+-- is exactly the path real instrumented code exercises.
+ioStackLevelOne :: (HasCallStack) => IO ()
+ioStackLevelOne = ioStackLevelTwo
+{-# NOINLINE ioStackLevelOne #-}
+
+ioStackLevelTwo :: (HasCallStack) => IO ()
+ioStackLevelTwo = error "boom"
+{-# NOINLINE ioStackLevelTwo #-}
+#endif
```
parsonsmatt: CPP here looks tricky to maintain in this file. I think there are tricks to making it a little easier to deal with- custom `src-dirs` in cabal file for certain ghc impls is a trick Gabby Gonzalez recommended to me once- but even just centralizing all the base dependent logic into single CPP blocks may be an improvement.

parsonsmatt (commented):
Implementation looks good to me, though I do have some future wins that I'd love to see

- Handling other kinds of exception annotations would be great, but *probably* requires passing in a `SomeExceptionAnnotation -> Span -> Span` that would let people do casts and modifications at will. Similar problem to the `annotated-exception` one + normal exception cast special handling.
- If the stock API (`recordException`) can *not* delete annotations, that'd be great, though I understand the GHC design decisions here are problematic.

api/src/OpenTelemetry/Trace/Core.hs
michaelpj: I totally agree with you that the `base` behaviour is wrong here, but I don't know if we can get around it. Although this function is somewhat problematic in any case:
1. If the user passes a `SomeException`, we'll strip the exception context (BAD)
2. If the user passes an `ExceptionWithContext`, we'll strip the exception context (BAD)
3. If the user passes a specific exception type... they've already lost the exception context (BAD)

I think the only solution is not having a function that accepts any `Exception`. I've added `recordSomeException` here, I think if we added `recordExceptionWithContext` as well, then I think that would actually cover us quite well:
- If you've got a `SomeException`, you'll use `recordSomeException` (GOOD), but if you do `recordExceptionWithContext (toException se)` that will also work (GOOD)
- If you want to work a specific exception type, then either:
    1. You try to use `recordExceptionWithContext`, you realise you don't have it, go and change the catch type so you have got it (GOOD)
    2. You can't change the catch site, so you use `recordSomeException (toException e)` (OK, we literally can't get the context unless you can change the catch site)

Then we should probably actively deprecate `recordException`.

api/test/OpenTelemetry/Trace/ExceptionHandlerSpec.hs
michaelpj: Happy to take a steer from @iand675 here. I do think we wan the CPP since this is new enough that we need to support the old way, but valuable enough that I do really want it when we can get it!

api/src/OpenTelemetry/Trace/Core.hs
michaelpj: I would prefer not to deal with this right now? For exceptions I think the story is pretty clear: there's a standard attribute, it's pretty clear how to render them, etc. Other annotations are... more complex. And I think we'd need some kind of user-pluggable thing here like you suggest.

Also - I don't currently _use_ other exception annotations, so I don't feel like I have a handle on what I'd want as a user!

api/src/OpenTelemetry/Trace/Core.hs
michaelpj: I've implemented this.

It looks like we might get a non-context discarding `toException` in a later base, but that'll be a while for most of us, so I think this is a good move still.

api/test/OpenTelemetry/Trace/ExceptionHandlerSpec.hs
michaelpj: I've isolated the CPP into an internal module. To make that work I added a stub `ExceptionWithContext` for earlier `base` versions with an unsatisfiable `Exception` instance, so people get okay error messages.

michaelpj:
Ready for another look I think

PR #312, opened 2026-06-24
Automatically initialize grpc trace exporter when OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc

jmorag:
2 commits:
1. 83a47fee9b92f1c32b5044cd98c1f9996fc3f700: very little functionality change, just makes the span exporter module import the internal config module.
2. 233e73eb563964d092486241853c66456b352dfe: make the default `otlpExporter` call the grpc exporter when the `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` config is set to `grpc`.


jmorag:
Added a few more commits to do the same grpc wiring for the metrics and logs exporters as well.

exporters/otlp/src/OpenTelemetry/Exporter/OTLP/LogRecord.hs
```diff
@@ -60,7 +67,17 @@ httpProtobufMimeType = "application/x-protobuf"
 
 
 otlpLogRecordExporter :: (MonadIO m) => OTLPExporterConfig -> m LogRecordExporter
-otlpLogRecordExporter conf = liftIO $ do
+#ifdef GRPC_ENABLED
```
michaelpj, line 70: The CPP is spreading out a bit, which is unfortunate. It's really better to corral it as much as possible, which I think we could do. The main thing we would need to do is make `grpcOtlpLogRecordExporter` etc. defined _unconditionally. Then this function can also unconditionally select it if it's requested.

The question is just: what should the value of the exporter be if we don't have `GRPC_ENABLED`? I can see two options:
1. A no-op exporter, possibly with an `unsafePerformIO` log line. A bit ugly, but we already do something like this in a few places.
2. `error`. Also ugly, but it should fail at startup, which seems like an okay time to crash if there are configuration issues.

Any thoughts @iand675 ?

